Issue #23, opened 2020-11-08
Experts: experts grid component


SergeOlabin:
timebox: 2d

Provide grid component for experts view

Issue #13, opened 2020-11-05
Header: create markup and navigation


SergeOlabin:
timebox: 2d

We need a header section so that user will be able to use app navigation and perform logic.

Don't focus on a markup elements too much since we don't have mocks yet. But, there should be semantic elements provided.

AC:
- [x] Header consists of 2 parts: 
- Icon, Logo, Auth
- NavBar
- [x] There're 2 parts of the header visible
- [x] Create navbar 
- [x] Navbar should perform routing

PR #36, opened 2020-11-12
Issue 23 experts grid component

aleksandra04:
develop

implemented css grid

## GitHub Board

**Issue link**

[#23 experts grid component ]( )

**Story link**

[#13 Experts]()

## Code reviewers
- [ ] @github_username

## Summary of issue

- [ ]  Summary of issue
- [ ]  Summary of issue

## Summary of change

- [ ]  Summary of change
- [ ]  Summary of change


src/modules/experts/styles/ExpertBlock.style.ts
```diff
@@ -0,0 +1,42 @@
...
+  },
+  item_8: {
+    gridArea: '1 / 5 / 3 / 6',
+  },
+  item_9: {
+    gridArea: '3 / 5 / 5 / 6',
+  },
+  item_10: {
+    gridArea: '1 / 6 / 5 / 7',
+  },
```
SergeOlabin, line 36: these should not be set in the `ExpertBlock`.

pass the styles from parent grid component, it should be responsible for styling it's children.
item doesn't have to know about it's position, it has to display the content, and that's it.
parent should be responsible for positioning its items.

src/lib/components/ExpertBlock.tsx
```diff
@@ -1,15 +1,24 @@
 import React from 'react';
+import { useStyles } from '../../modules/experts/styles/ExpertBlock.style';
 import { IExpert } from '../types';
 
 interface IExpertProps {
   expert: IExpert;
+  nameClass: string;
```
SergeOlabin, line 7: see another comment, it should not be here

SergeOlabin (changes requested):

src/modules/experts/styles/ExpertBlock.style.ts
aleksandra04: moved to MainExpertsView block

src/lib/components/ExpertBlock.tsx
aleksandra04: deleted

SergeOlabin (approved):

Issue #11, opened 2020-11-05
Newest: Lazy loading/pagination grid view


SergeOlabin:
timebox: 4d

![image](https://user-images.githubusercontent.com/22084409/98223997-93f93580-1f5b-11eb-8c02-f3ae36a7373e.png)

AC:
- [x] _NewestContainer_ component to manipulate data
- [x] generic (?) or 3rd party lib component to render data in the grid
- [x] _NewestContainer_ - [Render Props](https://reactjs.org/docs/render-props.html) and data manipulation
- [x] use redux for data fetching

PR #62, opened 2020-11-20
Issue 61 fix expert popover

DelFlumen:
develop

## GitHub Board

**Issue link**

[#11 Issue name ]( )

**Story link**

[#11 Story]( )

## Code reviewers
- [ ] @github_username

## Summary of issue

- [ ]  Summary of issue
- [ ]  Summary of issue

## Summary of change

- [ ]  Summary of change
- [ ]  Summary of change


src/lib/components/ExpertDataCard.tsx
```diff
@@ -8,17 +8,18 @@ import { DIRECTION_PROPERTIES } from './PostPreview/direction-properties';
 import { useStyles } from '../styles/ExpertDataCard.styles';
 
 export interface IExpertDataCardProps {
-  expert: IExpert;
+  expert: IExpert | null;
```
SergeOlabin, line 11: `ExpertDataCard` should not have `expert` as optional.
there's no point to render full info for an expert if there's no expert :)

instead, you can validate the expert on the higher level and render the `ExpertDataCard` only if `expert` is defined.

src/modules/main/components/MainExpertsView.tsx
```diff
@@ -30,26 +37,28 @@ export const MainExpertsView: React.FC = () => {
   const allExperts = cards.map((card, key) => (
     <>
       <div
-        onMouseEnter={handlePopoverOpen}
+        onMouseEnter={(event) => handlePopoverOpen(event, card)}
```
SergeOlabin, line 40: you can search the possibility for getting the expert or card from the event itselft, so that it'd be possible to reduce the number of arguments.

but, anyway, what it is now, it's ok.

SergeOlabin (changes requested):

src/lib/components/ExpertDataCard.tsx
DelFlumen: fixed

src/modules/main/components/MainExpertsView.tsx
DelFlumen: I searched but couldn't find a way to receive the card from the event itself.

src/lib/components/ExpertDataCard.tsx
```diff
@@ -14,11 +14,12 @@ export interface IExpertDataCardProps {
 const ExpertDataCard: React.FC<IExpertDataCardProps> = (props) => {
   const { expert } = props;
 
-  const fullName = `${expert.firstName} ${expert.secondName}`;
+  const fullName = expert && `${expert.firstName} ${expert.secondName}`;
```
SergeOlabin, line 17: remove `expert && ` conditions everywhere since `expert` is required now

SergeOlabin (changes requested):

src/lib/components/ExpertDataCard.tsx
DelFlumen: fixed

Issue #88, opened 2020-11-26
Direction: create state slice


calster00:
### Task description
Create state slice for Direction view

### To do 
1. - [x] Set up initial state using existing API types
2. - [x] Use Redux Toolkit to generate action creators & reducers

PR #89, opened 2020-11-27
Direction slice

darya42185:
## GitHub Board

[Direction: create state slice #88](https://github.com/ita-social-projects/dokazovi-fe/issues/88 )

## Code reviewers
- [ ] @github_username

## Summary of issue
1. - [x] Set up initial state using existing API types
2. - [x] Use Redux Toolkit to generate action creators & reducers


src/modules/direction/store/directionSlice.tsx
```diff
@@ -0,0 +1,36 @@
...
+  reducers: {
+    loadCourses: (state, action: PayloadAction<ICourse[]>) => {
+      state.courses = action.payload;
+    },
+  },
+});
+
+export const { loadCourses } = directionSlice.actions;
+
+export default directionSlice.reducer;
```
calster00, line 27: this has to be a part of `rootReducer`

SergeOlabin (changes requested):
pls merge develop to local branch

src/lib/components/Header/ListItems.tsx
```diff
@@ -0,0 +1,56 @@
+import React from "react";
```
SergeOlabin, line 1: this file was removed before, wasn't it?

src/store/rootReducer.ts
```diff
@@ -1,8 +1,10 @@
 import { combineReducers } from 'redux';
+import directionReducer from '../modules/direction/store/directionSlice';
 import mainReducer from '../modules/main/store/mainSlice';
 
 const rootReducer = combineReducers({
   main: mainReducer,
+  direction: directionReducer,
```
SergeOlabin, line 7: we need following structure for direction: `{ [%DirectionsEnumKey%]: directionReducer }`, so the overall structure will be: 
```
{
  main: MainReducer,
  directions: {
    covid19: directionReducer,
    therapy: directionReducer,
    // ...
  }
```
}

SergeOlabin (changes requested):

src/lib/components/Header/ListItems.tsx
darya42185: if you delete these commits, then this will lead to conflicts, so I think we should not delete them

Issue #121, opened 2020-12-07
[BUG] Fix auto post loading on every page visit


calster00:
### Task description
Right now posts fetching action is triggered whenever we visit main page (Newest) or direction page (Materials). Instead we should load additional posts only when a user clicks 'load more' button. We should differentiate between initial and subsequent loads and handle both cases appropriately.

### To do 
1. - [ ] Check if we load posts for the first time. 
2. - [ ] Either use separate actions or perform this check in the current action creator.

PR #126, opened 2020-12-08
Newest: Fix auto post loading on every page visit

Gtes:
Main Page Newest  Posts FIX

## GitHub Board

**Issue link**

[#121 Fix auto post loading on every page visit ](https://github.com/ita-social-projects/dokazovi-fe/issues/121 )

## Summary of issue

- [x]  Check if we load posts for the first time.
- [x]  Either use separate actions or perform this check in the current action creator.

## Summary of change
- [x]  Added additional thunk for initial newest posts load

### DIFF
--- a/src/modules/main/components/NewestContainer.tsx
+++ b/src/modules/main/components/NewestContainer.tsx
@@ -2,7 +2,11 @@ import React, { useEffect, useRef } from 'react';
 import { useSelector, useDispatch } from 'react-redux';
 import { Container, Grid, Button, Typography } from '@material-ui/core';
 import { RootStateType } from '../../../store/rootReducer';
-import { fetchNewestPosts, IMainState } from '../store/mainSlice';
+import {
+  fetchNewestPosts,
+  IMainState,
+  initialNewestPosts,
+} from '../store/mainSlice';
 import { useStyles } from './styles/NewestContainer.style';
 import BorderBottom from '../../../lib/components/Border';
 import PostsList from '../../../lib/components/PostsList';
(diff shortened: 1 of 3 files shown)


SergeOlabin (approved):

SergeOlabin (changes requested):
pls resolve conflicts :)

src/modules/main/store/mainSlice.ts
```diff
@@ -136,6 +136,19 @@ export const fetchExperts = createAsyncThunk('main/loadExperts', async () => {
...
 
+export const fetchInitialNewestPosts = (): AppThunkType => async (
+  dispatch,
+  getState,
+) => {
+  const {
+    main: { newest },
+  } = getState();
+
+  if (newest.newestPosts.length === 0) {
```
SergeOlabin, line 147:
```suggestion
  if (!newest.newestPosts.length) {
```
:)

SergeOlabin (approved):

src/modules/main/store/mainSlice.ts
SergeOlabin: you can not fix that now, just letting you know that it's also possible

src/modules/main/components/NewestContainer.tsx
```diff
@@ -24,7 +29,7 @@ const NewestContainer: React.FC = () => {
   });
 
   useEffect(() => {
-    setNewest();
+    setNewestInitial();
```
calster00, line 32: additional thunk for initial load is redundant, we can check if there are posts right here :)
```
    if (newestPosts.length === 0) {
      setNewest();
    }
```

calster00 (changes requested):

src/modules/main/components/NewestContainer.tsx
Gtes: On one of the syncs, we have discussed using the separate thunk for the initial load.

Checking length in the actual component was the first solution but was rejected.

Issue #127, opened 2020-12-08
Direction: Rework fetchMaterials to use createAsyncThunk


calster00:
### Task description
Rework fetchMaterials to use createAsyncThunk. Utilizing createAsyncThunk allows for better promise handling and makes controlling loading status easier.

### To do 
1. - [x] ~~Move current thunk logic to createAsyncThunk's payloadCreator callback.~~
2. - [x] ~~Handle promise lifecycle action types in extra reducers to update state.~~
3. - [x] Use LoadingInfo in MaterialsContainer.
4. - [ ] ~~Move current route into redux store instead of passing it from a component as an arg.~~
5. - [ ] ~~Thunk execution is cancelled if the direction is changed.~~
5. - [x] Add extended loading status handling in separate reducer.

PR #128, opened 2020-12-09
Add materials loading status handling 

calster00:
## GitHub Board

**Issue link**

[#127 Direction: Rework fetchMaterials to use createAsyncThunk ](https://github.com/ita-social-projects/dokazovi-fe/issues/127 )

## Summary of issue
- [x] Add extended loading status handling in separate reducer.
- [x] Use LoadingInfo in MaterialsContainer.


SergeOlabin (commented):
are we planning to re write to async thunk after introducing router to the store so that we could access `direnctionName` on loading?

calster00:
> are we planning to re write to async thunk after introducing router to the store so that we could access `direnctionName` on loading?

We gave up on that idea, since it has turned out we can't access root state (where current route is stored) from `directionSlice extraReducers` – they only get `directions` state slice. That is why I update loading status manually using `setMaterialsLoadingStatus`.

calster00 (commented):
> are we planning to re write to async thunk after introducing router to the store so that we could access direnctionName on loading?

We gave up on that idea, since it has turned out we can't access root state (where current route is stored) from directionSlice extraReducers – they only get directions state slice. That is why I update loading status manually using setMaterialsLoadingStatus.

calster00 (commented):
added thunk for initial posts load, ready for merge

SergeOlabin (approved):

src/modules/direction/components/MaterialsContainer.tsx
```diff
@@ -25,49 +22,59 @@ const MaterialsContainer: React.FC<IMaterialsContainerProps> = ({
...
+  } = useSelector(
     (state: RootStateType) => state.directions[direction.name].materials,
   );
 
   const dispatch = useDispatch();
 
   const dispatchFetchAction = () => {
-    dispatch(setMaterialsLoadingStatus(direction));
+    dispatch(
+      setMaterialsLoadingStatus({
```
SergeOlabin, line 36: pls put it to `fetchMaterials` so that fetchMaterials will handle loading status each time.
you are doing the same for initial loading

src/modules/direction/store/directionSlice.ts
```diff
@@ -132,68 +161,98 @@ export const fetchExperts = (
...
+  const { posts } = getState().directions[direction.name].materials;
+
+  if (posts.length === 0) {
+    dispatch(
+      setMaterialsLoadingStatus({
+        direction,
+        status: LoadingStatusEnum.pending,
+      }),
+    );
+    dispatch(fetchMaterials(direction));
```
SergeOlabin, line 256: why do we set loading only in case these's no posts but don't fetch the posts themselves?

SergeOlabin (changes requested):
pls resolve conflicts

Issue #70, opened 2020-11-24
Direction - filters for materials


SergeOlabin:
**Story link**
https://github.com/ita-social-projects/dokazovi-be/issues/22

### Task description
Filters section for the materials

### To do 
 - [x] https://github.com/ita-social-projects/dokazovi-fe/issues/69 is DONE
 - [x] use https://react-hook-form.com/ if possible
 - [x] GET request for entity types (Стаття, Допис ...) 
 - [x] GET request for tags
 - [x] use axios for GET requests 
 - [x] use redux to perform an action, store and get the data to the components.
 FILTERS:
 - [x] post types and tags are treated as a form components 
 - [x] on form state change - send a request to GET the updated data.
 - [x] filters are used as query params.
 - [x] use debounceTime for form state change - 500ms

### Estimation
[PERT-Sprint 2](https://docs.google.com/spreadsheets/d/14mClU39kUN25wvr-Di8eQHsuGhzPKMSsjcoCICIQJzU/edit#gid=212800467)

PR #129, opened 2020-12-09
filter for direction materials (types)

tania833:
develop

## GitHub Board

**Issue link**

[#70 ]


SergeOlabin (changes requested):
discussed on sync what to do

src/lib/constants/post-type-properties.ts
```diff
@@ -1,18 +1,22 @@
 
 export type PostTypePropertiesType = {
```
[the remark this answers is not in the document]
tania833: done

src/modules/direction/components/DirectionView.tsx
```diff
@@ -1,4 +1,5 @@
 import React, { useEffect } from 'react';
+import _ from 'lodash';
```
SergeOlabin, line 2: looks like this is not used

src/modules/direction/components/PostTypesFilter.tsx
```diff
@@ -0,0 +1,86 @@
...
+    (state: RootStateType) => state.properties.postTypes,
+  );
+
+  const state = () => {
+    const initialTypeState = {};
+    postTypes?.forEach((type) => {
+      initialTypeState[type.id.toString()] = false;
+    });
+    return initialTypeState;
+  };
```
SergeOlabin, line 31: this is `reduce` structure :)
```
const variable = {};
someArr.forEach(entry => variable[entry.id] = doSmth(entry));
return varuable;
```

```suggestion
  const state = () => postTypes?.reduce((acc, next) => acc[next.id.toString()], {}));
```

src/modules/direction/store/directionSlice.ts
```diff
@@ -31,6 +33,9 @@ export interface IDirectionState {
     meta: IExpertsMeta;
   };
   materials: IMaterialsState;
+  filters?: {
+    [key in DirectionFilterTypes]?: IFilter;
```
SergeOlabin, line 37: should this be optional?

src/modules/direction/store/directionSlice.ts
```diff
@@ -161,6 +166,25 @@ export const directionsSlice = createSlice({
...
+    ) => {
+      const { key, filters, directionName } = action.payload;
+      const direction = state[directionName] as IDirectionState;
+      if (direction) {
+        direction.filters = {
+          ...direction.filters,
+          [key]: filters,
+        };
+        direction.materials.meta.pageNumber = -1;
+        direction.materials.posts.length = 0;
```
SergeOlabin, line 185: should we also reset the posts?

SergeOlabin (changes requested):

src/modules/direction/store/directionSlice.ts
tania833: when filters object is empty

src/modules/direction/store/directionSlice.ts
tania833: yes, because now we have fetchInitialMaterials that checks if posts.length = 0

src/modules/direction/components/DirectionView.tsx
tania833: saw it in code swells and deleted

src/modules/direction/components/PostTypesFilter.tsx
tania833: done

SergeOlabin (approved):

Issue #114, opened 2020-12-07
Experts List: Expert Presentation Component


Gtes:
**Story link**
https://github.com/ita-social-projects/dokazovi-be/issues/90

### Task description
Each Experts card should contain:
- Author's picture
- First name, last name
- Specializations
- Place of work
- Position
- Last added material title (if exist)

![Screenshot](https://user-images.githubusercontent.com/71768856/100628075-7d62b600-3330-11eb-9335-175c89a4f47e.png)

### To do 
1. - [x] Create separate component for Expert card
1. - [x] Rough markup - to display all the fields necessary.

### Estimation
[PERT](https://docs.google.com/spreadsheets/d/14mClU39kUN25wvr-Di8eQHsuGhzPKMSsjcoCICIQJzU/edit#gid=1596292899)

PR #90, opened 2020-11-29
Direction course branch

darya42185:
develop

## GitHub Board

https://github.com/ita-social-projects/dokazovi-fe

[#11 Issue name ]( )

https://github.com/ita-social-projects/dokazovi-fe/issues/71

[#11 Story]( )

## Code reviewers
- [ ] @github_username

## Summary of issue

- [ ]  Summary of issue
- [ ]  Summary of issue

## Summary of change

- [ ]  Summary of change
- [ ]  Summary of change

PR #143, opened 2020-12-11
Issue #114 experts page card

AntonGluschuk:
Experts List: Expert Presentation Component

## GitHub Board

**Issue link**

[Experts List: Expert Presentation Component #114](https://github.com/ita-social-projects/dokazovi-fe/issues/114 )

**Story link**

[As a user(any role) I want to view Experts list on "Experts" page #90](https://github.com/ita-social-projects/dokazovi-be/issues/90)

## Summary of issue

- [x]  Create separate component for Expert card


src/lib/components/ExpertPhotoDataCard.tsx
```diff
@@ -0,0 +1,84 @@
+import React from 'react';
```
SergeOlabin, line 1: can we use a composition of 2 data cards for experts?
1 - card to display expert photo (used on main for example)
2- card to display expert description (used on main expert popover)

here we can use the composition of 1 and 2 ..

SergeOlabin (changes requested):

src/lib/components/ExpertPhotoDataCard.tsx
AntonGluschuk: done

SergeOlabin (approved):

Issue #119, opened 2020-12-07
Expert Profile: Expert Info


Gtes:
**Story link**
https://github.com/ita-social-projects/dokazovi-be/issues/94

### Task description
Show information about the Expert.

Expert’s profile:

- Author's picture
- First name, last name
- Position
- Specializations
- Place of work
- Phone number
- Email
- Info about expert

![](https://user-images.githubusercontent.com/71768856/100727953-97070a80-33cf-11eb-84e6-9a6b5032e935.png)

### To do 
- [x] Get info about Expert
- [x] use axios to perform the request
- [x] use redux to perform an action to fetch and store the data
- [x] Click on direction’s value should redirect to "Direction page"
- [x] Add spinner to display Loading - use [LoadingInfo.tsx](https://github.com/ita-social-projects/dokazovi-fe/blob/develop/src/lib/components/LoadingInfo.tsx) if possible

### Estimation
[PERT](https://docs.google.com/spreadsheets/d/14mClU39kUN25wvr-Di8eQHsuGhzPKMSsjcoCICIQJzU/edit#gid=1596292899)

PR #144, opened 2020-12-15
Expert Profile: Expert Info

calster00:
## GitHub Board

**Issue link**

[Expert Profile: Expert Info #119](https://github.com/ita-social-projects/dokazovi-fe/issues/119 )

## Summary of issue
Show information about the Expert:
- [x] Get info about Expert
- [x] use axios to perform the request
- [x] use redux to perform an action to fetch and store the data
- [x] Click on direction’s value should redirect to "Direction page"
- [x] Add spinner to display Loading - use LoadingInfo.tsx


src/lib/components/PostDirectionChip.tsx
```diff
@@ -20,9 +22,24 @@ const useStyles = makeStyles(
...
+  const history = useHistory();
+
+  const handleClick = () => {
+    const direction = Object.values(DIRECTION_PROPERTIES).find(
+      (d) => d.label === labelName,
+    );
+    if (direction?.route) {
+      history.push(`/direction/${direction.route}`);
+    }
+  };
```
SergeOlabin, line 34: putting a router logic in the chip is not a good idea...
chip is a basic UI component that can have a click handling logic implemented on a higher lvl, so that the component that uses chip will handle the click and, therefore, the routing part.

Since this chip is used ONLY for the direction, we can have this for now, I suppose.. But, bear in mind that you will need chip in another part of the app, so that you will have to separate the routing and representation logic of it. 
It's not a big deal to do that, but please be aware that you will have to do that when necessary :)

also, if putting the router inside the chip (since it's **direction** chip only), its props may be 
```
{
  direction: IDirection;
}
``` 
instead. So you won't have to get all the directions and searchBy label.

Also, search by label is bad, since `label` fields usually are not unique and if you do the search, it's better to search by id or something.

please, pass the direction as a prop to the chip.

SergeOlabin (changes requested):

src/lib/components/PostDirectionChip.tsx
calster00: I had some second thoughts about this part too, but left it as is until further discussion on daily.

> bear in mind that you will need chip in another part of the app, so that you will have to separate the routing and representation logic of it.

Good point, didn't notice that right away. I'll make a wrapper that adds interactive behavior then.

calster00: fixed!

SergeOlabin (approved):

PR #152, opened 2020-12-17
fix formatting

calster00:
Formatting fix, no logic changes.

### DIFF
--- a/src/App.test.tsx
+++ b/src/App.test.tsx
@@ -7,8 +7,8 @@ import { store } from './store/store';
 test('renders learn react link', () => {
   render(
     <Provider store={store}>
-    <App />
-    </Provider>
+      <App />
+    </Provider>,
   );
   expect(true).toBeTruthy();
 });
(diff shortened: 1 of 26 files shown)


SergeOlabin (approved):
cool!
did you use some command to do the formatting across the project?

if so, do you mind adding that command running on pre-push?

calster00:
@SergeOlabin I ran `prettier-format:all` from our `package.json`. It's complicated to do this pre-push, since after formatting you would also need to automatically make new commit (or amend last one) to save changes before pushing. 
I can add pre-commit hook instead to format staged files. It's not costly operation and only takes a second.

calster00 (commented):
@SergeOlabin I ran `prettier-format:all` from our `package.json`. It's complicated to do this pre-push, since after formatting you would also need to automatically make new commit (or amend last one) to save changes before pushing. 
I can add pre-commit hook instead to format staged files. It's not costly operation and only takes a second.

SergeOlabin:
@calster00 Seems good to me!
but pls discuss on the next sync with the team to get their opinion also

Issue #147, opened 2020-12-16
Adjust container sizes


SergeOlabin:
### Task description
Align container sizes.
They should have some min height value.

### AC
1. - [x] Spinners in the containers should be centered
2. - [x] Containers should have some min value
3. - [x] Containers should not collapse when having no data and a spinner
4. - [x] Direction page should not "jump" top when checking filters (if it's not about the issue with containers, create another task)

PR #162, opened 2020-12-21
Adjust container sizes

darya42185:
develop

## GitHub Board
https://github.com/ita-social-projects/dokazovi-fe/issues/147

[#147 Adjust container sizes ]( )

**Story link**

## Code reviewers
- [ ] @github_username

## Summary of issue

- [ ]  Summary of issue
- [ ]  Summary of issue

## Summary of change

- [ ]  Summary of change
- [ ]  Summary of change

### DIFF
--- a/src/modules/experts/components/ExpertsView.tsx
+++ b/src/modules/experts/components/ExpertsView.tsx
@@ -1,4 +1,3 @@
-/* eslint-disable @typescript-eslint/no-unsafe-assignment */
 import React, { useEffect } from 'react';
 import { Container, Grid } from '@material-ui/core';
 import { useSelector, useDispatch } from 'react-redux';
--- a/src/modules/experts/components/ExpertsView.tsx
+++ b/src/modules/experts/components/ExpertsView.tsx
@@ -1,3 +1,4 @@
+/* eslint-disable @typescript-eslint/no-unsafe-assignment */
 import React, { useEffect } from 'react';
 import { Container, Grid } from '@material-ui/core';
 import { useSelector, useDispatch } from 'react-redux';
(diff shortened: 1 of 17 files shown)


src/modules/experts/components/ExpertsView.tsx
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable @typescript-eslint/no-unsafe-assignment */
```
AntonGluschuk, line 1: Is this line necessary?

AntonGluschuk:
Awesome work!

Also ExpertProfileView.tsx need adjustments and that's all

src/modules/main/components/__test__/NewestContainer.spec.tsx
```diff
@@ -44,7 +44,7 @@ describe('newest', () => {
   it('should set loading state on succeeded when API call is pending', async () => {
     await store.dispatch(fetchNewestPosts());
 
-    expect(store.getState().main.newest.meta.loading).toEqual('succeeded');
+    // expect(store.getState().main.newest.meta.loading).toEqual('succeeded');
```
calster00, line 47: why disabling this? does loading status update correctly?

AntonGluschuk: Perhaps test fail through the deprecation of ```mainDirection``` field.
Thus after refactor to ```directions``` we would enable this test in future PR's.

tania833: Had the same problem, so agree

calster00 (approved):

SergeOlabin (approved):
looks nice.
but it's better to use flex for centering not the absolute positioning since, it better works when positioning in relation to the parent.
try avoiding using `position` property as much as you can, since it relies on the parent having some other positions as well.

discuss the comments with the team and can be merged.

Issue #120, opened 2020-12-07
Expert Profile: Filter for Materials


Gtes:
**Story link**
https://github.com/ita-social-projects/dokazovi-be/issues/94

### Task description
Filter for Expert Materials

### To do 
 - [ ] #118  IS DONE .
 - [ ] use https://react-hook-form.com/ if possible
 - [ ] GET request for entity types (Стаття, Допис ...) 
 - [ ] use axios for GET requests 
 - [ ] use redux to perform an action, store and get the data to the components.
 FILTERS:
 - [ ] post types  are treated as a form components 
 - [ ] on form state change - send a request to GET the updated data.
 - [ ] filters are used as query params.
 - [ ] use debounceTime for form state change - 500ms

### Estimation
[PERT](https://docs.google.com/spreadsheets/d/14mClU39kUN25wvr-Di8eQHsuGhzPKMSsjcoCICIQJzU/edit#gid=1596292899)

PR #166, opened 2020-12-22
Post types filter + post directions

calster00:
## GitHub Board

**Issue link**

[#120 Expert Profile: Filter for Materials](https://github.com/ita-social-projects/dokazovi-fe/issues/120 )

## Summary of issue

- [x]  Add post types filter for materials section

## Summary of change

- Changed post types to allow for multiple directions
- Fixed post cards to fit multiple directions
- Fixed repeated expert fetch
- Fixed filters not being reset when user leaves the page
- Move filters to their respective containers


src/modules/direction/components/DirectionView.tsx
```diff
@@ -46,21 +46,30 @@ const DirectionView: React.FC<IDirectionViewProps> = () => {
...
+      dispatch(
+        setPostFilters({
+          key: FilterTypeEnum.TAGS,
+          filters: {
+            value: [],
+          },
+          directionName: currentDirection.name,
+        }),
+      );
+    };
```
SergeOlabin, line 62: resetting the filters when changing the view, right?

do you think it's necessary?

SergeOlabin (approved):

SergeOlabin:
good stuff 👍

src/modules/direction/components/DirectionView.tsx
calster00: yes. that seemed reasonable, works like that on most webpages. 
checkboxes and tags inputs were unset after changing views, meaning their displayed state didn't match the one saved in redux, so we needed to either reset filters when leaving page or fix filters not displaying correct values when we visit the page again.

DelFlumen (approved):

Issue #116, opened 2020-12-07
Experts List: Regions Filter


Gtes:
**Story link**
https://github.com/ita-social-projects/dokazovi-be/issues/90

### Task description
Regions Filter section for the Experts

“Region”(“Регіон”) section should contain:

- List of regions
- Possibility of filtering with checkbox

![](https://user-images.githubusercontent.com/71768856/100630325-1692cc00-3333-11eb-95e2-96cf39b9ddb6.png)
![](https://user-images.githubusercontent.com/71768856/100630369-24e0e800-3333-11eb-866f-41eb62a54cae.png)

Click on \/ button should the full region's section come out.
Click on /\ button should the full region's section come back.

### To do 
 - [x] #115  is DONE
 - [x] use https://react-hook-form.com/ if possible
 - [x] GET request for Regions
 - [x] use axios for GET requests 
 - [x] use redux to perform an action, store and get the data to the components.
 - [x] Show/Hide all regions arrow
 - [x] Selected Regions should be displayed at the top.
 - [x] Selected Region should be removed from the top when user deselect it.
 FILTERS:
 - [x] Regions are treated as a form components 
 - [x] on form state change - send a request to GET the updated data.
 - [x] filters are used as query params.
 - [x] use debounceTime for form state change - 500ms

### Estimation
[PERT-Sprint 2](https://docs.google.com/spreadsheets/d/14mClU39kUN25wvr-Di8eQHsuGhzPKMSsjcoCICIQJzU/edit#gid=212800467)

PR #174, opened 2020-12-24
116 experts list filter fix

DelFlumen:
develop

## GitHub Board

**Issue link**

[#116 ]( )

## Summary of change

- [+]  Created FilterForm with logic and rough markup
- [+]  Added FilterForm to ExpertsView


src/lib/components/FilterForm.tsx
```diff
@@ -0,0 +1,132 @@
...
+  [key: string]: boolean;
+};
+
+export interface IInitLocalState {
+  [key: string]: boolean;
+}
+
+export interface ICheckboxes {
+  [key: number]: boolean;
+}
```
calster00, line 22:
```suggestion
export interface ICheckboxes {
  [key: string]: boolean;
}
```
you could use just one interface here, with string keys to match initial state

src/modules/experts/store/expertsSlice.ts
```diff
@@ -122,14 +134,28 @@ export const expertsSlice = createSlice({
...
+    setExpertsRegionsFilter: (state, action: PayloadAction<IFilter>) => {
+      if (state.experts.filters) {
+        state.experts.filters.REGIONS = action.payload;
+      }
+    },
+    setExpertsDirectionsFilter: (state, action: PayloadAction<IFilter>) => {
+      const value = action.payload;
+      if (state.experts.filters) {
+        state.experts.filters.DIRECTIONS = {
+          ...state.experts.filters.DIRECTIONS,
```
calster00, line 146: can't see what you spread here. shouln't this reducer be the same as `setExpertsRegionsFilter`?

src/modules/experts/store/expertsSlice.ts
```diff
@@ -68,7 +72,14 @@ const initialState: IExpertsState = {
       loading: LoadingStatusEnum.idle,
       error: null,
     },
-    filters: [],
+    filters: {
+      [FilterTypeEnum.DIRECTIONS]: {
+        value: [],
+      },
+      [FilterTypeEnum.REGIONS]: {
+        value: [],
```
calster00, line 80:
```suggestion
        value: {},
      },
      [FilterTypeEnum.REGIONS]: {
        value: {},
```

package.json
```diff
@@ -11,6 +11,7 @@
     "@testing-library/react": "^11.1.0",
     "@testing-library/user-event": "^12.1.10",
     "@types/lodash": "^4.14.165",
+    "@types/lodash.debounce": "^4.0.6",
```
SergeOlabin, line 14: it should already be present in "@types/lodash

src/modules/experts/store/expertsSlice.ts
SergeOlabin: seems like we are using an array of string values as a `value`, isn't it?

SergeOlabin (commented):
pls check the changes with @calster00 
also, check package-json

src/modules/experts/store/expertsSlice.ts
calster00: in `FilterForm` we assign an initial checkbox state of type `{ [key: string]: boolean; }` and later call `setExpertsRegionsFilter` with that value

src/modules/experts/store/expertsSlice.ts
DelFlumen: It's only a draft, these reducers will be the same. In this PR only regions filter is ready.

src/lib/components/FilterForm.tsx
DelFlumen: fixed

package.json
DelFlumen: fixed

SergeOlabin (approved):

darya42185 (approved):

Issue #176, opened 2020-12-28
Make experts list's FilterForm reusable


DelFlumen:
**Story link**
[#90 Story](https://github.com/ita-social-projects/dokazovi-be/issues/90)

### Task description
Make experts list's FilterForm reusable.

### To do 
1. - [x] Make experts list's FilterForm reusable.
2. - [x] Apply it to ExpertsView for Directions and Regions Filters.

PR #185, opened 2020-12-30
Issue 176 reusable experts list filter

DelFlumen:
develop

## GitHub Board

**Issue link**

[#176  Make experts list's FilterForm reusable ](https://github.com/ita-social-projects/dokazovi-fe/issues/176)

**Story link**

[#90 Story](https://github.com/ita-social-projects/dokazovi-be/issues/90)

## Summary of change

- [+] make experts list's FilterForm reusable;
- [+] apply it to ExpertsView for Directions and Regions Filters;
- [+] add accordion to FilterForm.


src/lib/components/FilterForm.tsx
```diff
@@ -44,7 +56,7 @@ export const FilterForm: React.FC<IFilterFormProps> = () => {
   const setExpertsFilter = useCallback(
     _.debounce((checkedTypes) => {
```
calster00, line 57:
```suggestion
    _.debounce((checkedTypes: ICheckboxes) => {
```
use proper function typing instead of assertion to make sure function calls are correctly typed

AntonGluschuk (approved):

src/lib/components/FilterForm.tsx
```diff
@@ -44,7 +56,7 @@ export const FilterForm: React.FC<IFilterFormProps> = () => {
   const setExpertsFilter = useCallback(
```
calster00, line 56: this is fine just now, but consider moving this callback (as well as the one in effect hook) to the parent component, so that this filter form doesn't dictate the type of reducer arguments, as it might change in the future

src/modules/direction/store/directionSlice.ts
```diff
@@ -291,7 +291,7 @@ export const fetchMaterials = (direction: IDirection): AppThunkType => async (
 
     const response = await getPosts('latest-by-direction', {
       params: {
-        direction: direction.id,
+        direction: +direction.id,
```
calster00, line 294: this one is already a number, see IDirection
```suggestion
        direction: direction.id,
```

calster00 (changes requested):

src/lib/components/FilterForm.tsx
```diff
@@ -2,27 +2,36 @@ import React, { useCallback, useEffect } from 'react';
...
 }
 
-export interface IFilterFormProps {}
+export interface IObjForAction {
+  value: ICheckboxes;
+}
+
+export interface IFilterFormProps {
+  setExpertsListFilter: (obj: IObjForAction) => void;
+  expertsProperties: IPostType[];
```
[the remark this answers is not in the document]
DelFlumen: fixed

src/lib/components/FilterForm.tsx
DelFlumen: fixed

src/modules/direction/store/directionSlice.ts
DelFlumen: fixed

calster00 (approved):

Issue #191, opened 2021-01-11
Auth: modals


SergeOlabin:
**Story link**
[#100 Story](https://github.com/ita-social-projects/dokazovi-be/issues/29)

### Task description
Login via UI.
we need login and registration modals.

### To do 
1. - [x] UI modals for authentification

Q:
- useForm?

PR #208, opened 2021-01-20
Issue 191 login modal

DelFlumen:
develop

## GitHub Board

**Issue link**

[#191 Auth: modals](github.com/ita-social-projects/dokazovi-fe/issues/191)

## Summary of change

- [x]  created LoginModal
- [x]  created RegistrationModal
- [x]  connected RegistrationModal to LoginModal
- [x]  add LoginModal to Header


src/lib/components/Modals/LoginModal.tsx
```diff
@@ -0,0 +1,196 @@
...
+  // eslint-disable-next-line @typescript-eslint/unbound-method
+  const { register, handleSubmit, errors } = useForm<IInputs>();
+
+  const handleLoginOpen = () => {
+    setLoginOpen(true);
+  };
+
+  const handleLoginClose = () => {
+    setLoginOpen(false);
+  };
```
SergeOlabin, line 35: you can avoid using such methods.. 
just use `setLoginOpen(true)` when you need that, no need for wraping this :)

src/lib/components/Modals/LoginModal.tsx
```diff
@@ -0,0 +1,196 @@
...
+      <Dialog
+        open={loginOpen}
+        onClose={handleLoginClose}
+        aria-labelledby="form-dialog-title"
+      >
+        <DialogTitle id="form-dialog-title">
+          Введіть Ваші email та пароль
+          {/* {response.error && (
+                  <span style={{ color: 'red' }}>Неправильний email або пароль</span>
+                )} */}
```
SergeOlabin, line 81: remove

src/lib/components/Modals/LoginModal.tsx
```diff
@@ -0,0 +1,196 @@
...
+                    required: {
+                      value: true,
+                      message: "Це поле є обов'язковим",
+                    },
+                    pattern: {
+                      value: /^[A-Z0-9._%+-]+@[A-Z0-9.-]+\.[A-Z]{2,}$/i,
+                      message:
+                        'Email повинен містити символ @ та одну крапку після нього',
+                    },
+                  })}
```
SergeOlabin, line 102: I'd suggest you to move these configs out of `return (` statement, that'll be easier to manage them latter

```
{
                    required: {
                      value: true,
                      message: "Це поле є обов'язковим",
                    },
                    pattern: {
                      value: /^[A-Z0-9._%+-]+@[A-Z0-9.-]+\.[A-Z]{2,}$/i,
                      message:
                        'Email повинен містити символ @ та одну крапку після нього',
                    },
                  }
```
that pieces that are iniside the register function

SergeOlabin (approved):
very nice!

check the comments, those are recomendations, you can ignore or stick to them, up to you

src/lib/components/Modals/LoginModal.tsx
DelFlumen: This part will be uncommented when there will be connect with BE: it's a message in case of bad response from server (email and/or password turn out to be wrong).

src/lib/components/Modals/LoginModal.tsx
DelFlumen: Should I move it to separate object and then apply to register(), using spread? Like: `register(...emailValidationConfig)`

PR #230, opened 2021-02-02
fix: history push state

AntonGluschuk:
Fix history pushState on ArticleCreation and NoteCreation

## Summary of change

- [x]  Send newPost object to the preview instead of the func


src/modules/postCreation/PostCreationPreview.tsx
```diff
@@ -50,11 +52,20 @@ const PostCreationPreview: React.FC = () => {
...
   };
 
+  const sendPost = async () => {
+    const responsePost = await postPublishPost(currentState.publishPost);
+    history.push(`/posts/${responsePost.data.id}`);
+  };
+
+  const publishNewPost = () => {
+    sendPost();
+  };
```
SergeOlabin, line 62: what is the reason wrapping this ? :)

AntonGluschuk: fixed

Gtes (approved):

DelFlumen (approved):

Issue #274, opened 2021-02-23
Video: preview


SergeOlabin:
[SKIP] until we figure out what post video format will be 
### Task description
after Create Video page is ready, 

### To do 
1. - [x] video preview card on the Create Video page
2. - [x] show video in the preview
3. - [x] show a description partitially

Ihor-malyshko:
this is part of the #395

PR #390, opened 2021-06-03
add iframe in video preview card

Ihor-malyshko:
Issue link

#274 Issue Video: preview card


sarhan-azizov (commented):
need to approve from BA

PR #478, opened 2021-07-29
#469/cover with tests header component

NabokinAlexandr:

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,6 +33,7 @@
     "react-slick": "^0.27.13",
     "react-toastify": "^7.0.3",
     "redux": "^4.1.0",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -33,6 +33,7 @@
     "react-slick": "^0.27.13",
     "react-toastify": "^7.0.3",
     "redux": "^4.1.0",
+    "redux-mock-store": "^1.5.4",
```
michalenr, line 36: I wouldn't recommend using this library, because it hasn't been main for 3+ years

Here's a good guide on testing redux from the official docs: https://redux.js.org/usage/writing-tests

PR #558, opened 2021-09-02
edit Important mobile view

0shar0:
[user story](https://github.com/ita-social-projects/dokazovi-requirements/issues/206)

### DIFF
--- a/src/old/provider/MobileProvider/ScreenProvider.tsx
+++ b/src/old/provider/MobileProvider/ScreenProvider.tsx
@@ -4,8 +4,8 @@ import { ScreenContext } from './ScreenContext';
 
 export const ScreenProvider: React.FC = ({ children }) => {
   const theme = useTheme();
-  const mobile = useMediaQuery(theme.breakpoints.down('sm'));
-  const tablet = useMediaQuery(theme.breakpoints.down('md'));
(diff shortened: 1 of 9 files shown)


src/old/provider/MobileProvider/ScreenProvider.tsx
```diff
@@ -4,8 +4,8 @@ import { ScreenContext } from './ScreenContext';
 
 export const ScreenProvider: React.FC = ({ children }) => {
   const theme = useTheme();
-  const mobile = useMediaQuery(theme.breakpoints.down('sm'));
-  const tablet = useMediaQuery(theme.breakpoints.down('md'));
+  const mobile = !useMediaQuery(theme.breakpoints.up('sm'));
```
michalenr, line 7: It is a convention in JS to use `is`, `has`, etc. prefixes for boolean values.

So, `isMobile` should be better.

michalenr: Also, what was the problem with using `.down('sm)` ?

0shar0: it should works like on this image
Screenshot_6


but with this solution it works like this
Screenshot_3

michalenr (approved):

niksonax (approved):

PR #780, opened 2022-09-06
fix(docker):  extend copy options in docker

Dergelyova:
### Type of Pull Request *

 ### Description *
Extend copy of files in docker.prod file. Now lock files copied and used in build


skyboyer (approved):
nice. before we were copying just package.json which caused `yarn install` to operate without lock file, which was copied later. It caused dependencies were never locked during the build and finally(yesterday) caused build failure, when installed latest prettier-format incompatible with Node12 project uses.

PR #787, opened 2022-09-22
bug 563: update date formating and related tests

anechytailo:
Pull Request Template
 

### Type of Pull Request *
- [x] ENHANCEMENT  (non-breaking change which improves existing functionality)

### Related links
 
**Issue link**

 
**Story link**
[#563 Story](https://github.com/ita-social-projects/dokazovi-requirements/issues/563)
 
### Description *
Date/time of the published cards is not displayed in full format on the material cards
-> `formatDate` function's been updated
 
### Summary of change
Material Cards date/time formating
 
 
 
"*" - mandatory field


anechytailo:
The main update is in `src/utilities/formatDate.ts` file. Everything else is styling and test updates.

src/utilities/formatDate.ts
```diff
@@ -1,22 +1,37 @@
...
+  '11': i18n.t(langTokens.date.novemberGenitiveCase),
+  '12': i18n.t(langTokens.date.decemberGenitiveCase),
+};
 
-  return day && month ? `${+day.toString()} ${monthNames[+month]}` : '';
+export const formatDate = (dateString = '', time = false): string => {
+  if (dateString.length) {
+    const fullDate = new Date(dateString);
+    const hours =
+      fullDate.getHours() < 10
```
skyboyer, line 22: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/padStart is well supported:

fullDate.getMinutes().toString().padStart(2, '0')

src/components/Posts/PostInfo/PostInfo.styles.ts
```diff
@@ -5,7 +5,7 @@ export const useStyles = makeStyles(
     root: {
       display: 'flex',
       marginBottom: theme.spacing(8),
-      [theme.breakpoints.down('xs')]: {
+      [theme.breakpoints.down('sm')]: {
```
skyboyer, line 8: why styles are changing? I don't get that from story 563

src/utilities/formatDate.ts
anechytailo: riiight! did think of this approach :(
thanks! working on it!

src/components/Posts/PostInfo/PostInfo.styles.ts
anechytailo: I slightly updated the moment when we should split the single line into two to avoid dramatic layout breaking. It happened because new date format `hh:mm, dd MM yyyy` much longer than previous `dd.mm.yyyy`

skyboyer: understood, thank you for clarification

src/utilities/formatDate.ts
anechytailo: @skyboyer please review the updates

src/utilities/formatDate.ts
```diff
@@ -1,22 +1,31 @@
...
+  '7': i18n.t(langTokens.date.julyGenitiveCase),
+  '8': i18n.t(langTokens.date.augustGenitiveCase),
+  '9': i18n.t(langTokens.date.septemberGenitiveCase),
+  '10': i18n.t(langTokens.date.octoberGenitiveCase),
+  '11': i18n.t(langTokens.date.novemberGenitiveCase),
+  '12': i18n.t(langTokens.date.decemberGenitiveCase),
+};
 
-  return day && month ? `${+day.toString()} ${monthNames[+month]}` : '';
+export const formatDate = (dateString = '', time = false): string => {
```
skyboyer, line 18: side note: it's fine by now, but if there more formats are introduced, we better consider delegating this to special library instead of implementing by ourselves

skyboyer (approved):

vpavlikk (approved):

PR #553, opened 2021-08-29
Filters in Materials page fixed

NabokinAlexandr:

PR #790, opened 2022-10-10
Issues 553 554 555

vpavlikk:
Pull Request Template
 

### Type of Pull Request *

### Related links
 
**Issue link**
[#553 “Список авторів” table]( https://github.com/ita-social-projects/dokazovi-requirements/issues/553 )
[#554 Author search]( https://github.com/ita-social-projects/dokazovi-requirements/issues/554 )
[#555 “Список авторів” page]( https://github.com/ita-social-projects/dokazovi-requirements/issues/555 )
 
### Description *
Describe the big picture of your changes here.
If it fixes a bug or resolves a feature request, be sure to link to that issue.
 
###Summary of issue
Describe the issue
 
###Summary of change
Describe the solution
 
 
 
"*" - mandatory field


src/models/experts/reducers.ts
```diff
@@ -1,19 +1,33 @@
...
     expertIds: [],
     experts: {},
-    meta: {
-      totalPages: 0,
-      pageNumber: 0,
-      isLastPage: false,
-      totalElements: 0,
+    totalPages: 0,
+    isLastPage: true,
+    totalElements: 0,
```
skyboyer, line 20: why this information has been moved from `meta`? I understand that "total experts number" and "current page" are _different_ metadata, but still it's not like specific expert's name or region(which are definitely `data`)

skyboyer: or is this changed due to changes on backend side?

src/views/Profile/AuthorsList/AuthorsListFilters.tsx
```diff
@@ -0,0 +1,50 @@
...
+  const [boundedSetSize, boundedSetField] = useActions([setSize, setField]);
+
+  const handleNotesToShowChange = (val) => {
+    boundedSetSize(val);
+  };
+
+  return (
+    <Box className={classes.filtersBox}>
+      <Typography variant="body1" className={classes.notesToShowPanel}>
+        {i18.t(langTokens.admin.showNotesAmount).split(' ')[0]}
```
skyboyer, line 32: this is probably better to be split into separate `admin.showNotesAmountLeft` and `admin.showNotesAmountRight`

splitting by " " is not flexible enough, and we may want to have several words to the left. But splitting by spaces and picking just first word will not allow that

src/models/experts/asyncActions.ts
```diff
@@ -47,13 +47,69 @@ export const fetchExperts = createAsyncThunk(
...
+          page,
+          size: Number(size),
+          sort: isSearchByAuthorexists(author),
+          userName: author,
         },
+      });
+
+      response.data.content.forEach((expert) => {
+        expert['isAllowedToDelete'] = expert.postStatuses.every(
+          (postStatus) => postStatus.status === 'DRAFT',
```
skyboyer, line 95: so if there is publication in `MODERATION_SECOND_SIGN` we cannot delete author. But is it right? Not saying that code does wrong, but may we double check with BA/PO?

src/old/lib/utilities/API/api.ts
```diff
@@ -303,3 +303,7 @@ export const updatePlatformInformation = async (
 ): Promise<AxiosResponse<PlatformInformationType>> => {
   return instance.put(`/platform-information/`, requestBody);
 };
+
+export const getDoctorsList = async (): Promise<AxiosResponse> => {
```
skyboyer, line 307: but why without subtype for `AxiosResponse<...>`? and where is it used?

src/views/Profile/AdminTable/AdminTextField.tsx
```diff
@@ -2,9 +2,10 @@ import React from 'react';
 import FormControl from '@material-ui/core/FormControl';
 import TextField from '@material-ui/core/TextField';
 import { FieldEnum, IField } from '../../../models/adminLab/types';
+import { TextFieldsEnum } from '../../../models/experts/types';
 
 interface IMaterialsTextField {
-  field: FieldEnum;
+  field: FieldEnum | TextFieldsEnum;
```
skyboyer, line 8: I don't see `TestFieldsEnum` in https://github.com/ita-social-projects/dokazovi-fe/blob/epic_14_Authors_list_and_personal_info/src/models/experts/types.ts

what do I miss?

src/views/Profile/AuthorsList/AuthorsList.tsx
```diff
@@ -1,113 +1,95 @@
...
-    {
-      id: 6,
-      firstName: 'Таржеман',
-      lastName: 'Соколов',
-      region: 'Дніпропетровська область',
-      dateOfCreation: '12.03.2022',
-      dateOfEdition: '12.03.2022',
-      bio:
-        'Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor ',
+  const maxCouldBePerPage = (pageNumber + 1) * Number(size);
```
skyboyer, line 45: sorry, I'm kind of lost here

may you describe what's logic behind `from` and `to` by few examples?

src/views/Profile/AuthorsList/AuthorsTableBody.tsx
```diff
@@ -32,12 +50,27 @@ const AuthorsTableBody: React.FC<IAuthorsTableBodyProps> = ({ authors }) => {
...
+        <TableCell>
+          <div className={classes.modifSection}>
+            <IconButton
+              aria-label="edit profile"
+              onClick={() => handleChangeClick(id)}
+              className={classes.editButton}
+            >
+              <EditOutlined />
+            </IconButton>
+            <AuthorsDeletingDialog
```
skyboyer, line 67: I think it will be much more flexible, readable and better for testing if we put conditional rendering here:

```
{isAllowedToDelete ? 
  <AuthorsDeletingDialog id={id} fullName={fullName} /> :
  <AuthorCannotBeDeletedTooltip fullName={fullName} />
}
```

src/views/Profile/AuthorsList/AuthorsDeletingDialog.tsx
```diff
@@ -0,0 +1,104 @@
...
+        setOpenTooltip(value);
+        break;
+      default:
+        break;
+    }
+  };
+
+  return (
+    <>
+      <Tooltip
```
skyboyer, line 44: I think it's better not to merge 2 unrelated component into one. Commented in  AuthorsTableBody component's code

src/views/Profile/AuthorsList/AuthorsDeletingDialog.tsx
```diff
@@ -0,0 +1,104 @@
...
+        open={openTooltip && !isAllowedToDelete}
+        onClose={() => handleChanges('tooltip', false)}
+        onOpen={() => handleChanges('tooltip', true)}
+        placement="bottom-end"
+        title={
+          'Дозволяється видаляти тільки тих авторів, які не мають постів або всі пости мають статус "Чернетка"!'
```
skyboyer, line 50: i18n :)

src/models/adminLab/types.ts
```diff
@@ -111,10 +111,10 @@ export enum StatusesForActions {
 }
 
 export interface IAdminLabExpert extends IExpert {
-  region?: string;
```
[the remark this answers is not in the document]
vpavlikk: I just moved it to IExpert type

src/models/experts/asyncActions.ts
```diff
@@ -47,13 +47,69 @@ export const fetchExperts = createAsyncThunk(
...
+            size,
+            textFields: { author },
+            sort: { order, sortBy },
+          },
+        },
+      } = getState() as any;
+
+      const isSearchByAuthorexists = (arg: string) => {
+        if (!arg) {
+          return `${sortBy.split(',')}` + `,${order}`;
```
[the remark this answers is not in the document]
vpavlikk: when we sort by fullName, on the server side the is no such key. Only firstName/lastName exist. In this case we should combine two keys for one column fullName and while sending the request pass multiple values for one parameter. This could be soloved by passing  array of param values [val1, val2] - that is the reason of using split()

src/models/experts/asyncActions.ts
```diff
@@ -47,13 +47,69 @@ export const fetchExperts = createAsyncThunk(
...
+        }
+        return;
+      };
+
+      const { page } = options;
+      const response = await getAllExperts({
+        params: {
+          page,
+          size: Number(size),
```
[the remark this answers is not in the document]
vpavlikk: removed unnecessary method Number.It was added by mistake, the type of size is always number, and this key has default value 25

src/models/experts/selectors.ts
```diff
@@ -1,15 +1,17 @@
...
+export const selectExpertsMeta = (state: RootStateType): IExpertMeta =>
+  state.experts.meta;
 
 export const selectExpertsLoading = (state: RootStateType): LoadingStatusEnum =>
   state.experts.loading;
 
-export const selectExpertById = (state: RootStateType, id: number): IExpert => {
-  return store.getState().experts.data.experts[id];
-};
+export const selectExpertsByIds = (ids: number[]): IExpert[] =>
```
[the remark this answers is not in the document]
vpavlikk: the were two selectors which returned the same result and  and took data from the slice of state in the same way. I just deleted duplicate

src/models/experts/types.ts
```diff
@@ -1,27 +1,67 @@
...
+  totalPages: number;
+  isLastPage: boolean;
+  totalElements: number;
 }
 
-export interface IExpertsState {
-  data: IExpertsData;
-  loading: LoadingStatusEnum;
-  error: null | string;
+export enum AutorsListSortBy {
```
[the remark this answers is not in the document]
vpavlikk: done

src/models/experts/types.ts
```diff
@@ -1,27 +1,67 @@
...
+}
+
+export enum AutorsListOrder {
+  desc = 'desc',
+  asc = 'asc',
+}
+
+export interface ISortAutorsList {
+  order: keyof typeof AutorsListOrder;
+  sortBy: string;
```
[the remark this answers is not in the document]
vpavlikk: no, because keys and values of enum AuthorsListSortBy are not the same for 'region' and 'city'

src/models/experts/asyncActions.ts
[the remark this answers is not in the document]
skyboyer: so sortBy can be: 'firstName', 'lastName' or 'fullName' and for last one we should pass to the server 'firstName,lastName'. Is it correct? If yes, and we always append `,${order}` can we just ``return `${sortBy},${order}`;`` ? 

Ideally, I'd expect some mapping on UI side, like 

```
sortTypesMap = {
  firstName: ['firstName'],
  fullName: ['firstName', 'lastName']
}
```

so we could have search presets instead of hardcoding, but this is rather low-value-higher-efforts thing. So I'm not sure if it makes sense even to discuss by now.

src/views/Profile/AuthorsList/AuthorListDropdown.tsx
```diff
@@ -0,0 +1,41 @@
+import React from 'react';
+import { FormControl, Select, InputBase, MenuItem } from '@material-ui/core';
+import { useStyles } from './styles/AuthorListDropdown.styles';
+
+interface IAuthorListDropdownProps {
+  options: number[];
```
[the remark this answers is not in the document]
vpavlikk: yes, it is for limiting max experts per page. options = [10, 25, 50, 100]

src/models/experts/asyncActions.ts
```diff
@@ -47,13 +47,69 @@ export const fetchExperts = createAsyncThunk(
...
+          meta: {
+            size,
+            textFields: { author },
+            sort: { order, sortBy },
+          },
+        },
+      } = getState() as any;
+
+      const isSearchByAuthorexists = (arg: string) => {
+        if (!arg) {
```
skyboyer, line 77: separate question: why to do check for `arg`? I understand it as "if we search for specific author we don't pass sorting argument". But Is it really necessary? If there is just single match, then it does not matter if we sort or not. And if there are more than 1 match found, we definitely better to have results ordered by gives field. Or does backend reject having both filtering and sorting?

src/views/Profile/AuthorsList/AuthorListDropdown.tsx
```diff
@@ -0,0 +1,41 @@
...
+  const classes = useStyles();
+  const handleChange = (event: React.ChangeEvent<{ value: unknown }>) => {
+    setChanges(event.target.value as number);
+  };
+
+  return (
+    <>
+      <FormControl variant="standard">
+        <Select
+          labelId="demo-customized-select-label"
```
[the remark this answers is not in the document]
vpavlikk: the lines were unnecessary, removed them

src/views/Profile/AuthorsList/styles/AuthorsTable.styles.tsx
```diff
@@ -0,0 +1,63 @@
...
+    paddingTop: '0px',
+  },
+
+  dialogContent: {
+    padding: '0px 0px',
+  },
+
+  dialogContentText: {
+    '& .MuiDialogContent-root': {
+      padding: '0px 0px 0px 0px',
```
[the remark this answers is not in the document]
vpavlikk: sorry, just have not noticed it, yes i understand that 0 and '0px 0px 0px 0px' cause to the same effect ))

src/models/experts/asyncActions.ts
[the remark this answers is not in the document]
vpavlikk: done

src/views/Profile/AuthorsList/styles/AuthorsListFilters.styles.ts
```diff
@@ -0,0 +1,21 @@
...
+
+    '& .MuiTextField-root': {
+      margin: theme.spacing(1),
+      padding: theme.spacing(1),
+      backgroundColor: theme.palette.common.white,
+      borderRadius: '2px',
+    },
+  },
+  notesToShowPanel: {
+    margin: 'auto 15px auto auto',
```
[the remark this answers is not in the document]
vpavlikk: no, used for element centration

src/views/Profile/AuthorsList/AuthorListDropdown.tsx
```diff
@@ -0,0 +1,41 @@
...
+}
+
+export const AuthorListDropdown: React.FC<IAuthorListDropdownProps> = ({
+  options,
+  selected,
+  setChanges,
+}) => {
+  const classes = useStyles();
+  const handleChange = (event: React.ChangeEvent<{ value: unknown }>) => {
+    setChanges(event.target.value as number);
```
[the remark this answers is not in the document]
vpavlikk: rewrote the lines  event: React.ChangeEvent<{ value: number | unknown }>, and setChanges(event.target.value);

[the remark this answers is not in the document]
vpavlikk: "as number just makes typescript shut up and thinking it will be a number" - Yes

src/models/experts/asyncActions.ts
vpavlikk: it is a temporary solution because of the bug on the server side while having both filtering and sorting

src/views/Profile/AuthorsList/AuthorsTableBody.tsx
```diff
@@ -1,29 +1,47 @@
...
+
+    const parsedCreatedAt = format(parseISO(createdAt as string), 'd.MM.yyyy');
+    const parsedEditedAt = editedAt
+      ? format(parseISO(editedAt), 'd.MM.yyyy')
+      : parsedCreatedAt;
+
     const fullName = `${firstName} ${lastName}`;
+
+    const handleChangeClick = (arg: number) => {
+      alert(`author${arg} changing toggled`);
```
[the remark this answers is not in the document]
vpavlikk: has removed

src/old/lib/utilities/API/api.ts
vpavlikk: unnecessary method, have already removed it

src/views/Profile/AuthorsList/AuthorsDeletingDialog.tsx
vpavlikk: done

src/models/experts/index.ts
```diff
@@ -1,8 +1,16 @@
...
+  expertsReducer,
+  setSize,
+  setPageNumber,
+  setField,
+  setSort,
+  setExpertsStateToInit,
+} from './reducers';
+export {
+  selectExpertsData,
+  selectExpertsMeta,
```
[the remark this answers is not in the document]
vpavlikk: Screenshot 2022-10-11 at 21 57 40

src/views/Profile/AdminTable/AdminTextField.tsx
vpavlikk: Screenshot 2022-10-12 at 00 04 12

src/views/Profile/AuthorsList/AuthorsListFilters.tsx
vpavlikk: done

src/old/modules/experts/components/Accordion.tsx
```diff
@@ -77,31 +77,34 @@ const Accordion: React.FC<IAccordion> = ({ expert }) => {
...
+                  key={socialNetwork}
                 >
-                  {expert.socialNetwork}
-                </NavLink>
-              </Typography>
-            </div>
-          ) : null}
+                  <img
+                    className={classes.contactIcons}
+                    src={facebook}
```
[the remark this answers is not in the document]
vpavlikk: Is not in scope of current task, the the changes were made in the file only because of adding new fields into type IExpert

src/old/modules/experts/components/Accordion.tsx
```diff
@@ -77,31 +77,34 @@ const Accordion: React.FC<IAccordion> = ({ expert }) => {
...
-              />
-              <Typography variant="body1" className={classes.contacts}>
-                <NavLink
-                  to={{
-                    pathname: `${expert.socialNetwork}`,
-                  }}
-                  target="_blank"
-                  className={classes.links}
+          {expert?.socialNetworks && expert.socialNetworks.length > 0
+            ? expert.socialNetworks.map((socialNetwork) => (
```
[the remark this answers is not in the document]
vpavlikk: done

src/views/Profile/AuthorsList/AuthorsTableBody.tsx
```diff
@@ -1,29 +1,47 @@
...
       region,
-      dateOfCreation,
-      dateOfEdition,
+      createdAt,
+      editedAt,
+      mainInstitution,
+      isAllowedToDelete,
     } = author;
+
+    const parsedCreatedAt = format(parseISO(createdAt as string), 'd.MM.yyyy');
```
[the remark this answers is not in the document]
vpavlikk: done

src/views/Profile/AuthorsList/AuthorsTableBody.tsx
```diff
@@ -32,12 +50,27 @@ const AuthorsTableBody: React.FC<IAuthorsTableBodyProps> = ({ authors }) => {
             {fullName}
           </Typography>
         </TableCell>
-        <TableCell>{region}</TableCell>
-        <TableCell>City</TableCell>
-        <TableCell>{dateOfCreation}</TableCell>
+        <TableCell>{region?.name}</TableCell>
```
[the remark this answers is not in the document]
vpavlikk: temporary solution. I added as an optional because as i understood for some of the test authors, fields 'region' and 'main Institution' equal to null. But while creating an author using basic logic such fields are required and they will always have valid values

src/views/Profile/AuthorsList/AuthorsList.tsx
vpavlikk: Screenshot 2022-10-13 at 12 43 37

src/models/experts/asyncActions.ts
vpavlikk: the question has been already discussed with BA. For now author deleting should been processed on the way specified in the issue https://github.com/ita-social-projects/dokazovi-requirements/issues/553 - 
The delete icon should appear only near the Autor who:
a) Don’t have any materials;
b) All materials are in status “Чернетка”.

src/views/Profile/AuthorsList/AuthorsList.tsx
vpavlikk: Screenshot 2022-10-13 at 17 26 35

src/old/modules/experts/components/Accordion.tsx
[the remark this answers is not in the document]
skyboyer: understood

src/models/experts/reducers.ts
```diff
@@ -1,19 +1,33 @@
...
   data: {
     expertIds: [],
     experts: {},
-    meta: {
-      totalPages: 0,
-      pageNumber: 0,
-      isLastPage: false,
-      totalElements: 0,
+    totalPages: 0,
+    isLastPage: true,
```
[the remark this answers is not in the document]
vpavlikk: yes, if the current page will be the last one of all existed pages

src/views/Profile/AuthorsList/AuthorsDeleteBtn.tsx
```diff
@@ -0,0 +1,45 @@
...
+  const handleOpenTooltip = () => {
+    setOpenTooltip(true);
+  };
+  const handleCloseTooltip = () => {
+    setOpenTooltip(false);
+  };
+
+  return (
+    <>
+      <Tooltip
```
skyboyer, line 27: I think we don't need separate state here. it looks like there is no _single_ prop to disable tooltip from appearing, but we can do it like:

```
<Tooltip
  disableFocusListener={isAllowedToDelete}
  disableHoverListener={isAllowedToDelete}
  disableTouchListener={isAllowedToDelete}
```

skyboyer: or even with passing empty `title`, api docs say that in this case tooltip will not be shown:

```
<Tooltip
  title={isAllowedToDelete ? null : tooltipTitle}
```

src/models/experts/asyncActions.ts
skyboyer: I don't insist, but maybe it does not really make sense to have separate function after logic has been simplified; as a bonus we could put comment on that in single place(with separate function it's harder to find right place for a comment):

```
sort: author ? undefined : `${sortBy},${order}` // backend crashes if both filtering and sorting are passed, see <link to backend issue>
```

src/models/experts/reducers.ts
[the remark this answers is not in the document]
skyboyer: then instead of saving extra flag we could just create separate selector:

```
isLastPageSelector = (pageNumber) => (state) => pageNumber === state.....totalPages - 1
```
and use it in component 

```
const currentPageNumber = useSelector ....
const isLastPage = useSelector(isLastPageSelector(currentPageNumber));
```

or 

```
isLastPageSelector = (state, page/number) => (state) => pageNumber === state.....totalPages - 1

```

with using as

```
const isLastPage = useSelector(state => isLastPageSelector(state, currentPageNumber));
```

what do you say?

src/views/Profile/AuthorsList/AuthorsList.tsx
skyboyer: I think it could be simplified by checking for `isLastPage`:

```
const rangeStart = pageNumber * pageSize;
const rangeEnd = isLastPage ? totalElements : (pageNumber + 1 ) * pageSize;
```

that's I've renamed `from` and `end` just for clarity; I'm not insistin on using that name

src/old/modules/experts/components/ExpertInfo.tsx
```diff
@@ -36,7 +36,7 @@ const ExpertInfo: React.FC<IExpertInfoProps> = ({ expert }) => {
           </Box>
         </Box>
         <Grid container item className={classes.accordionWrapper}>
-          {expert.email || expert.socialNetwork ? (
+          {expert.email || expert.socialNetworks.length > 0 ? (
```
skyboyer, line 39: nit: better keep consistency with the same check: either both do `.length > 0` or both do `.length !== 0`. Given that `length` cannot be negative, I'm rather for second version

src/views/Profile/AuthorsList/AuthorListDropdown.tsx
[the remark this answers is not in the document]
skyboyer: I see. If it does not need big efforts, I propose to name prop `pageSizeValues` or `pageSizes` instead of quite generic `options` name

src/views/Profile/AuthorsList/AuthorsListFilters.tsx
```diff
@@ -0,0 +1,54 @@
...
+  const [boundedSetSize, boundedSetField] = useActions([setSize, setField]);
+
+  const handleNotesToShowChange = (val) => {
+    boundedSetSize(val);
+  };
+
+  return (
+    <Box className={classes.filtersBox}>
+      <Typography variant="body1" className={classes.notesToShowPanel}>
+        <Trans
```
skyboyer, line 33: awesome.

src/views/Profile/AuthorsList/AuthorsTableHead.tsx
```diff
@@ -7,94 +7,105 @@ import {
...
-  //   };
+    if (sortKey === sortBy) {
+      newOrder =
+        order === AutorsListOrder.desc
+          ? AutorsListOrder.asc
+          : AutorsListOrder.desc;
+    }
+    boundedSetSort({
+      order: newOrder,
+      sortBy: sortKey as keyof typeof AuthorsListSortBy,
```
skyboyer, line 88: also "autors" without "h". Please check if there are other instances across files you changed anyway.

src/views/postCreation/PostCreation.tsx
```diff
@@ -304,7 +304,7 @@ export const PostCreation: React.FC<IPostCreationProps> = ({
         title: savedPostDraft.title,
         videoUrl: savedPostDraft.videoUrl,
         type: { id: postType.type, name: postType.name },
-      } as IPost),
+      } as unknown) as IPost),
```
skyboyer, line 307: why double type assertion was needed here?

vpavlikk: Screenshot 2022-10-13 at 22 29 40

src/old/modules/experts/components/ExpertInfo.tsx
vpavlikk: done

src/models/experts/asyncActions.ts
vpavlikk: Done

src/views/Profile/AuthorsList/AuthorListDropdown.tsx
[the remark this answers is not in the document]
vpavlikk: Done

src/models/experts/reducers.ts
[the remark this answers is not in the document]
vpavlikk: This also makes sense but in this case I should rewrite the logic in src/old/modules/experts/components/ExpertsView.tsx and src/old/modules/experts/components/ExpertMaterialsContainer.tsx as the component use the same slice of state as src/views/Profile/AuthorsList/AuthorsList.tsx and also remove isLastPage flag from the server side because the flag becomes unnecessary and there is no reason to get it in the response.

src/views/Profile/AuthorsList/AuthorsDeleteBtn.tsx
vpavlikk: changed to the first version described in comment

src/models/experts/reducers.ts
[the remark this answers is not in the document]
skyboyer: decided not to do it this way, since `isLastPage` comes from server

src/views/postCreation/PostCreation.tsx
skyboyer: ideally we might have `IPost` without `id` and `type ISavedPost = IPost & {  id: number }` but if that's a lot of changes, let's keep it as double conversion(but I would still consider this as a techdebt)

skyboyer: ...or in opposite: `IUnsavedPost = Omit<IPost, "id">`

src/views/Profile/AuthorsList/AuthorsTableHead.tsx
```diff
@@ -7,94 +7,105 @@ import {
...
-  //   };
+    if (sortKey === sortBy) {
+      newOrder =
+        order === AutorsListOrder.desc
+          ? AutorsListOrder.asc
+          : AutorsListOrder.desc;
+    }
+    boundedSetSort({
+      order: newOrder,
+      sortBy: sortKey as keyof typeof AutorsListSortBy,
```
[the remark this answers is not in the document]
skyboyer: does not give much value, closing

skyboyer (approved):

PR #791, opened 2022-10-17
improvement: update date formatting function to work with timestamps

anechytailo:
Pull Request Template
 

### Type of Pull Request *
- [x] ENHANCEMENT  (non-breaking change which improves existing functionality)

### Related links
 
**Issue link**

 
**Story link**

 
### Description *
Update Date formatting function, make it work with timestamps.
 
### Summary of issue
Describe the issue
 
### Summary of change
Describe the solution
 
 
 
"*" - mandatory field

### DIFF
--- a/src/utilities/formatDate.ts
+++ b/src/utilities/formatDate.ts
@@ -15,9 +15,9 @@ const monthNames: { [index: string]: string } = {
   '12': i18n.t(langTokens.date.decemberGenitiveCase),
 };
 
-export const formatDate = (dateString = '', time = false): string => {
-  if (dateString.length) {
-    const fullDate = new Date(dateString);
+export const formatDate = (dateValue = '', time = false): string => {
+  if (dateValue) {
+    const fullDate = new Date(dateValue);
     const hours = fullDate.getHours().toString().padStart(2, '0');
     const mins = fullDate.getMinutes().toString().padStart(2, '0');
     const displayDate = time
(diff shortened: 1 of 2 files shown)


src/utilities/formatDate.ts
```diff
@@ -15,9 +15,9 @@ const monthNames: { [index: string]: string } = {
   '12': i18n.t(langTokens.date.decemberGenitiveCase),
 };
 
-export const formatDate = (dateString = '', time = false): string => {
-  if (dateString.length) {
-    const fullDate = new Date(dateString);
+export const formatDate = (dateValue = '', time = false): string => {
```
skyboyer, line 18: but what is the type of dateValue? is not it only `String` without `number`?

anechytailo: added update

src/utilities/formatDate.ts
```diff
@@ -15,9 +15,12 @@ const monthNames: { [index: string]: string } = {
...
 };
 
-export const formatDate = (dateString = '', time = false): string => {
-  if (dateString.length) {
-    const fullDate = new Date(dateString);
+export const formatDate = (
+  dateValue: number | string = '',
+  time = false,
+): string => {
+  if (dateValue) {
```
skyboyer, line 22: oh, I checked only now that there is `date-fns` library in the project(https://date-fns.org/v2.29.3/docs/format)

it will not help with replacing `monthNames` in Ukrainian, but besides that we can consume it to format rest of the string:

```
format(fullDate, `kk:mm dd ${monthNames[fullDate.getMonth() + 1]} yyyy`)
```

anechytailo: thanks for the advice
updated in accordance

skyboyer (approved):
cool, looks good to me

vpavlikk (approved):

PR #802, opened 2022-12-17
Fix bug resolve conflict 799

artemhoriuk1994:


skyboyer (approved):
this is a bit weird to see fix for #563 to be merged into `develop` again 
maybe makes sense to ask QA to smoke-test specificially #563 to ensure it's still fine

otherwise I think we are good to merge this

PR #804, opened 2023-02-15
Issue 619 changes log

artemhoriuk1994:
Pull Request Template
 

### Type of Pull Request *
- [ x ] ENHANCEMENT  (non-breaking change which improves existing functionality)

#
 
### Description *
Did everything according to story #619. The option to export data to a .csv file remained unfulfilled.

image

### DIFF
--- a/src/models/changeLog/asyncAction.ts
+++ b/src/models/changeLog/asyncAction.ts
@@ -1,0 +1,15 @@
+import { createAsyncThunk } from '@reduxjs/toolkit';
+import { fetchMaterialsChange } from 'old/lib/utilities/API/api';
+import { IChangeLog } from './types';
+
+export const getChangeLog = createAsyncThunk(
+  'getChangeLogActions',
+  async (options: IChangeLog, { rejectWithValue }) => {
+    try {
+      const changes = await fetchMaterialsChange(options);
+      return changes.data;
+    } catch (error) {
+      return rejectWithValue(error.response.data);
+    }
+  },
+);
--- a/src/models/changeLog/asyncAction.ts
+++ b/src/models/changeLog/asyncAction.ts
@@ -1,12 +1,12 @@
 import { createAsyncThunk } from '@reduxjs/toolkit';
-import { fetchMaterialsChange } from 'old/lib/utilities/API/api';
-import { IChangeLog } from './types';
+import { fetchChangeLog } from 'old/lib/utilities/API/api';
+import { IChangeLogOptions } from './types';
 
 export const getChangeLog = createAsyncThunk(
   'getChangeLogActions',
-  async (options: IChangeLog, { rejectWithValue }) => {
+  async (options: IChangeLogOptions, { rejectWithValue }) => {
     try {
-      const { data } = await fetchMaterialsChange(options);
+      const { data } = await fetchChangeLog(options);
       return data;
     } catch (error) {
       return rejectWithValue(error);
--- a/src/models/changeLog/asyncAction.ts
+++ b/src/models/changeLog/asyncAction.ts
@@ -6,10 +6,10 @@ export const getChangeLog = createAsyncThunk(
   'getChangeLogActions',
   async (options: IChangeLog, { rejectWithValue }) => {
     try {
-      const changes = await fetchMaterialsChange(options);
-      return changes.data;
+      const { data } = await fetchMaterialsChange(options);
+      return data;
     } catch (error) {
-      return rejectWithValue(error.response.data);
+      return rejectWithValue(error);
     }
   },
 );
--- a/src/models/changeLog/selectors.ts
+++ b/src/models/changeLog/selectors.ts
@@ -1,5 +1,5 @@
 import { RootStateType } from 'models/rootReducer';
 
-export const selectSize = (state: RootStateType) => {
+export const selectSize = (state: RootStateType): number | undefined => {
   return state.changesLog.size;
 };
--- a/src/models/changeLog/selectors.ts
+++ b/src/models/changeLog/selectors.ts
@@ -1,5 +1,5 @@
 import { RootStateType } from 'models/rootReducer';
 
 export const selectSize = (state: RootStateType) => {
-  return state.changesLog;
+  return state.changesLog.size;
 };
--- a/src/models/changeLog/selectors.ts
+++ b/src/models/changeLog/selectors.ts
@@ -1,5 +1,4 @@
 import { RootStateType } from 'models/rootReducer';
-import { IChangeLog } from './types';
 
 export const selectSize = (state: RootStateType) => {
   return state.changesLog;
--- a/src/models/changeLog/selectors.ts
+++ b/src/models/changeLog/selectors.ts
@@ -1,0 +1,6 @@
+import { RootStateType } from 'models/rootReducer';
+import { IChangeLog } from './types';
+
+export const selectSize = (state: RootStateType) => {
+  return state.changesLog;
+};
--- a/src/old/lib/components/Chips/ChipsList.tsx
+++ b/src/old/lib/components/Chips/ChipsList.tsx
@@ -9,11 +9,7 @@ export interface IChipsListProps {
   max?: boolean;
 }
 
-const ChipsList: React.FC<IChipsListProps> = ({
-  checkedNames,
-  isLabelItem,
-  max,
-}) => {
+const ChipsList: React.FC<IChipsListProps> = ({ checkedNames }) => {
   const { directions } = useSelector(
     (state: RootStateType) => state.properties,
   );
--- a/src/old/lib/utilities/API/api.ts
+++ b/src/old/lib/utilities/API/api.ts
@@ -2,6 +2,7 @@
 import axios, { AxiosError, AxiosRequestConfig, AxiosResponse } from 'axios';
 import qs from 'qs';
 import { toast } from 'react-toastify';
+import { IChangeLog } from 'models/changeLog/types';
 import {
   CreatePostRequestUnionType,
   CreateTagRequestType,
@@ -27,11 +28,10 @@ import {
   ActiveDirectionType,
   PlatformInformationType,
   UpdatePlatformInformationRequestType,
-  GetConfigType,
+  GetChangeLogType,
 } from './types';
 import { BASE_URL } from '../../../apiURL';
 import { getToken } from '../../../provider/AuthProvider/getToken';
-import { IChangeLog } from 'models/changeLog/types';
 
 export const instance = axios.create({
   baseURL: BASE_URL,
@@ -310,8 +310,8 @@ export const getPlatformInformation = async (
 
 export const fetchMaterialsChange = async (
   parametres?: IChangeLog,
-): Promise<AxiosRequestConfig> => {
-  return await instance.get(`/log/post-logs`, {
+): Promise<AxiosResponse<GetChangeLogType>> => {
+  return instance.get(`/log/post-logs`, {
     params: {
       size: parametres?.size,
     },
--- a/src/old/lib/utilities/API/api.ts
+++ b/src/old/lib/utilities/API/api.ts
@@ -1,8 +1,13 @@
 /* eslint-disable no-param-reassign */
-import axios, { AxiosError, AxiosRequestConfig, AxiosResponse } from 'axios';
+import axios, {
+  AxiosError,
+  AxiosRequestConfig,
+  AxiosResponse,
+  AxiosPromise,
+} from 'axios';
 import qs from 'qs';
 import { toast } from 'react-toastify';
-import { IChangeLog } from 'models/changeLog/types';
+import { IChangeLogOptions } from 'models/changeLog/types';
 import {
   CreatePostRequestUnionType,
   CreateTagRequestType,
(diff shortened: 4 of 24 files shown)


src/old/lib/utilities/API/api.ts
```diff
@@ -306,6 +308,16 @@ export const getPlatformInformation = async (
   return instance.get(`/platform-information/${id}`);
 };
 
+export const fetchMaterialsChange = async (
```
skyboyer, line 311: I propose to name on `fetchChangeLog` or in some other way but to make `changeLog` to be part of the name

src/views/Profile/AdminTable/AdminDatePicker.tsx
```diff
@@ -65,7 +65,7 @@ export const AdminDatePicker: React.FC<IMaterialsDate> = ({
               onChange={(value) =>
                 setChanges({
                   option: 'start',
-                  date: value && value.toISOString(),
+                  date: value && value.toISOString().substring(0, 10),
```
skyboyer, line 68: I propose to introduce helper in /src/utilities/formatDate.ts for date formating

1. shorter code
2. can cover with tests(also for edge cases)
3. consistent formatting across different occurences

do you agree?

src/views/postUpdation/AllPostTypesUpdation.tsx
```diff
@@ -444,6 +479,31 @@ export const AllPostTypesUpdation: React.FC<IAllPostTypesUpdation> = ({
...
+        <AuthorListDropdown
+          pageSizes={changesPerPage}
+          setChanges={handleNotesToShowChange}
+          selected={
+            size === 0 ? changesPerPage[changesPerPage.length - 1] : size
+          }
+        />
+        <List>
+          {changes.length > 0 ? (
+            changes.map((item) => {
```
skyboyer, line 494: another approach is to render `changes.map` always:

```
<List>
  {changes.map(item => .....)}
  {changes.length === 0 && <>No Changes</>}
```

Definitely no difference in behaviour. But to me this variant is easier to follow, than ternary `? :`

src/views/postUpdation/AllPostTypesUpdation.tsx
```diff
@@ -444,6 +479,31 @@ export const AllPostTypesUpdation: React.FC<IAllPostTypesUpdation> = ({
...
         post={post}
       />
+      <BorderBottom />
+      <h1>Changes List</h1>
+      <Box mt={3}>
+        <AuthorListDropdown
+          pageSizes={changesPerPage}
+          setChanges={handleNotesToShowChange}
+          selected={
+            size === 0 ? changesPerPage[changesPerPage.length - 1] : size
```
skyboyer, line 489: let's go with `size || changesPerPage[changesPerPage.length - 1]`

not just shorter(code size should never be an argument per se) but also covers edge case with `undefined`(not sure whether it's allowed for this type but anyway)

src/views/postUpdation/AllPostTypesUpdation.tsx
```diff
@@ -444,6 +479,31 @@ export const AllPostTypesUpdation: React.FC<IAllPostTypesUpdation> = ({
...
+            changes.map((item) => {
+              return (
+                <ListItem key={item.id}>
+                  {item.title} {item.changes}:{' '}
+                  {new Date(item.dateOfChange).toUTCString()}
+                </ListItem>
+              );
+            })
+          ) : (
+            <>No Chanages</>
```
skyboyer, line 503: may we want i18n here?

src/views/postUpdation/AllPostTypesUpdation.tsx
```diff
@@ -444,6 +479,31 @@ export const AllPostTypesUpdation: React.FC<IAllPostTypesUpdation> = ({
         disabled={isDisabled}
         post={post}
       />
+      <BorderBottom />
+      <h1>Changes List</h1>
```
skyboyer, line 483: also, may we want to keep it in i18n strings?

src/models/changeLog/asyncAction.ts
```diff
@@ -0,0 +1,15 @@
+import { createAsyncThunk } from '@reduxjs/toolkit';
+import { fetchMaterialsChange } from 'old/lib/utilities/API/api';
+import { IChangeLog } from './types';
+
+export const getChangeLog = createAsyncThunk(
```
skyboyer, line 5: but where this action is used?

src/views/Profile/AdminTable/AdminTableBody.tsx
```diff
@@ -68,9 +69,11 @@ const AdminTableBody: React.FC<IAdminTableBodyProps> = ({ isAdmin }) => {
         <TableCell>{id}</TableCell>
         <TableCell className={classes.titleCol}>
           <Chip label={typeName} className={classes[chipClass] as string} />
-          <Typography variant="h6" component="p" className={classes.title}>
-            {title}
-          </Typography>
+          <Link to={`/edit-post?id=${id}`} className={classes.title}>
```
skyboyer, line 72: we cannot use this here, but just as knowledge sharing: there is `generatePath` helper in `react-router-dom`

so if `/edit-post/:id` was a named route(which it is NOT and `id` parameter in query part instead of string - but let's imagine) then we could:

```
import { EDIT_POST_ROUTE } from '../...../routes.js';

  <Link to={generatePath(EDIT_POST_ROUTE, { postId: id })}

```

but here we cannot use that trick, so just saying

src/models/changeLog/selectors.ts
```diff
@@ -0,0 +1,5 @@
+import { RootStateType } from 'models/rootReducer';
+
+export const selectSize = (state: RootStateType) => {
+  return state.changesLog;
```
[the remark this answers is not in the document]
artemhoriuk1994: so it should be state.changeLog.size
First, I got the size object through destructuring in the component
So im chnge that for more better readability

src/views/postUpdation/AllPostTypesUpdation.tsx
artemhoriuk1994: yeap im agree with you that this one way look more better

src/models/changeLog/asyncAction.ts
artemhoriuk1994: this acition im used in reducet like extraReducer 
```suggestion
const changeLog = createSlice({
  name: 'changesLog',
  initialState,
  reducers: {
    setChangesSize: (state, action: PayloadAction<number>) => {
      const setChangeObj = state;
      setChangeObj.size = action.payload;
    },
  },
  extraReducers: {
    getAsyncActionsReducer(getChangeLog as any),
  },
});
```

src/views/Profile/AdminTable/AdminDatePicker.tsx
artemhoriuk1994: Create function helper for this date request

src/old/lib/components/Chips/ChipsList.tsx
```diff
@@ -9,11 +9,7 @@ export interface IChipsListProps {
   max?: boolean;
 }
 
-const ChipsList: React.FC<IChipsListProps> = ({
-  checkedNames,
-  isLabelItem,
-  max,
-}) => {
+const ChipsList: React.FC<IChipsListProps> = ({ checkedNames }) => {
```
skyboyer, line 12: since we don't use `max` and `isLabelItem` anymore, may it make sense to remove them from `IChipsListProps` as well?

src/old/lib/utilities/API/types.ts
```diff
@@ -277,3 +278,36 @@ export type UpdatePlatformInformationRequestType = {
   title?: string;
   text?: string;
 };
+
+export type ChangeLogType<T = any> = {
```
skyboyer, line 282: seems `T` is not used in this generic, so do we need it at all here?

src/old/modules/experts/components/Accordion.tsx
```diff
@@ -1,5 +1,4 @@
 import React, { useState } from 'react';
-import { NavLink } from 'react-router-dom';
```
skyboyer, line 2: more cleanup! I like that!

src/styles/theme/index.ts
```diff
@@ -10,7 +10,7 @@ export const SCREEN_BREAKPOINTS = {
   xl: 1920,
 };
 
-export const MAIN_THEME = createMuiTheme({
+export const MAIN_THEME = createTheme({
```
skyboyer, line 13: yeeeey! no more deprecated function!

src/utilities/formatDate.ts
```diff
@@ -42,3 +42,7 @@ export const displayShortDate = (
   }
   return `${dateValue}`;
 };
+
+export const requestDate = (value: Date): string => {
+  return value && value.toISOString().substring(0, 10);
```
skyboyer, line 47: since `value` is declared as a `Date` there is no need to do `value &&`. we cannot pass `undefined` or `null`. right?

src/views/Profile/AuthorsList/AuthorListDropdown.tsx
```diff
@@ -3,9 +3,9 @@ import { FormControl, Select, InputBase, MenuItem } from '@material-ui/core';
 import { useStyles } from './styles/AuthorListDropdown.styles';
 
 interface IAuthorListDropdownProps {
-  pageSizes: number[];
-  selected: number;
-  setChanges: (payload: any) => void;
+  pageSizes: Array<number | string>;
```
skyboyer, line 6: I believe we should strict: `pageSizes: Array<number | 'All'>`; similarly to `selected` property.

src/views/Profile/AdminTable/AdminTableBody.tsx
```diff
@@ -68,9 +69,17 @@ const AdminTableBody: React.FC<IAdminTableBodyProps> = ({ isAdmin }) => {
         <TableCell>{id}</TableCell>
         <TableCell className={classes.titleCol}>
           <Chip label={typeName} className={classes[chipClass] as string} />
-          <Typography variant="h6" component="p" className={classes.title}>
-            {title}
-          </Typography>
+          <Link
+            to={{
+              pathname: '/edit-post',
```
skyboyer, line 74: cool! I like object-style target for `<Link>`

src/views/postUpdation/AllPostTypesUpdation.tsx
```diff
@@ -444,6 +478,34 @@ export const AllPostTypesUpdation: React.FC<IAllPostTypesUpdation> = ({
...
       />
+      <BorderBottom />
+      <Typography component="h2" variant="h2">
+        {t(langTokens.admin.changesList)}
+      </Typography>
+      <Box mt={3}>
+        <AuthorListDropdown
+          pageSizes={changesPerPage}
+          setChanges={handleNotesToShowChange}
+          selected={size || changesPerPage[changesPerPage.length - 1]}
```
skyboyer, line 489: if we update `IAuthorListDropdownProps` to allow only `All` among other strings, we could also hard-code it here:

```
selected={size || 'All'}
```
and TS will keep us safe

skyboyer (commented):
looks good, have few comments

src/old/lib/components/Chips/ChipsList.tsx
artemhoriuk1994: I think its  better to move `IChipsListProps` to a new file, cause another components use this interfece

src/utilities/formatDate.ts
artemhoriuk1994: yeap thats my fault 
im tottaly agree with u

src/views/Profile/AuthorsList/AuthorListDropdown.tsx
artemhoriuk1994: We have problem like this
image

skyboyer: give me some time, I'll play with it

src/views/Profile/AuthorsList/AuthorListDropdown.tsx
```diff
@@ -2,10 +2,12 @@ import React from 'react';
...
 import { useStyles } from './styles/AuthorListDropdown.styles';
 
+export type PageSizeType = Array<number | 'All'>;
+
 interface IAuthorListDropdownProps {
-  pageSizes: number[];
-  selected: number;
-  setChanges: (payload: any) => void;
+  pageSizes: PageSizeType;
+  selected: number | string | undefined;
```
skyboyer, line 9: but what if we had `type PageSizeType = number | 'All'`? Then we could type `selected: PageSizeType | undefined`

or does it create any issues in other code?

src/views/Profile/AuthorsList/AuthorListDropdown.tsx
```diff
@@ -2,10 +2,12 @@ import React from 'react';
 import { FormControl, Select, InputBase, MenuItem } from '@material-ui/core';
 import { useStyles } from './styles/AuthorListDropdown.styles';
 
+export type PageSizeType = Array<number | 'All'>;
```
skyboyer, line 5: as for naming, may it use plural like `PageSizesType` for an array?

skyboyer (approved):
looks good to me, just 2 minor comments

PR #805, opened 2023-02-15
Bugfix627 create author

dima-diachkov:
Pull Request Template
 

### Type of Pull Request *

### Related links
 
**Issue link**
[#627 Bug Report 'Адміністрування' page: The page for creating the new record is not opened on the site after clicking the 'Створити нового автора' button](https://github.com/ita-social-projects/dokazovi-requirements/issues/627 )
 
**Story link**
[#555 “Список авторів” page](https://github.com/ita-social-projects/dokazovi-requirements/issues/575 )
 
### Description *
done according to bug report #627. 
 
ezgif com-video-cutter (1)


src/views/Profile/PersonalInfo/PersonalInfo.tsx
```diff
@@ -1,116 +1,528 @@
...
+    // TODO dispatch new author to server
+  };
+
+  useEffect(() => {
+    toggleButtonHandler();
+  }, [errorFields]);
+
+  const toggleButtonHandler = () => {
+    /* eslint-disable-next-line */
+    for (const [_, value] of Object.entries(errorFields)) {
```
skyboyer, line 305: you can just `const [, value] of` to ignore value at 0 position

src/views/Profile/PersonalInfo/PersonalInfo.tsx
```diff
@@ -1,116 +1,528 @@
...
+  const [newAuthorValues, setNewAuthorValues] = useState<INewAuthorValues>({
+    avatar: '',
+    firstName: '',
+    lastName: '',
+    regionId: 0,
+    cityId: 0,
+    bio: '',
+    email: '',
+    /* eslint-disable-next-line */
+    socialNetwork: new Array(5).fill(null),
```
skyboyer, line 48: I think that `[null, null, null, null, null]` is better(for readability and also for ESLint - why it complains btw?) to provide literally(since 5 does not come from external constant but is hard-coded)

src/views/Profile/PersonalInfo/PersonalInfo.tsx
```diff
@@ -1,116 +1,528 @@
...
+    bio: false,
+    facebook: false,
+    instagram: false,
+    youtube: false,
+    twitter: false,
+    linkedin: false,
+  });
+  const [toggleButton, setToggleButton] = useState(true);
+
+  const setRegionByCityId = async (id: number) => {
```
skyboyer, line 82: may it make sense to move region picker with all this fetching thing into separate component?

src/views/Profile/PersonalInfo/PersonalInfo.tsx
```diff
@@ -1,116 +1,528 @@
...
+        ...errorFields,
+        [inputIdentifier]: `Можливо ввести максимум ${maxSymbols} символів`,
+      });
+      return false;
+    }
+    setErrorFields({ ...errorFields, [inputIdentifier]: '' });
+    return true;
+  };
+
+  const inputNameChangeHandler = (
```
skyboyer, line 137: I think we better group validation rules in separate config object:

```
const validation = {
  name: {
    min: 2,
    max: 30,
    allowed: fields.acceptedSymbols,
  },
  bio: {
    min: 3,
    max: 250,
   
  }
```


Also I'm wondering if we need regexp when we validate just specific characters, not combinations of them etc. So maybe it makes sense to configure string with allowed characters? And use it to both validation and for error message?

src/views/Profile/PersonalInfo/PersonalInfo.tsx
```diff
@@ -1,116 +1,528 @@
...
-import { IExpert } from '../../../old/lib/types';
-import { getCurrentUser } from '../../../old/lib/utilities/API/api';
-import { ERROR_404 } from '../../../old/lib/constants/routes';
+import { ICity, IRegion } from '../../../old/lib/types';
+import { ErrorField } from './ErrorField';
+import { regex } from './constants/regex';
+import { fields } from './constants/fields';
+import { INewAuthorValues } from './types';
 
 export const PersonalInfo: React.FC = () => {
```
skyboyer, line 29: I think this component has grown too big - not just in lines count but in responsibility. It renders fields, handles changes, does validation, shows errors and loads regions. Too much for single component. Ideally it should handle input and keep local state but delegate validation and should not load regions at all.

Propose to reshape it with extracting region selector into separate component and moving validation logic in separate config object(probably keeping it in the same file, but just out of the component)

what do you say?

src/views/Profile/PersonalInfo/RegionCityHandler.tsx
```diff
@@ -0,0 +1,172 @@
...
+    });
+    if (regionResponse) {
+      setSelectRegions([regionResponse]);
+      setRegionForValue(regionResponse);
+    }
+  };
+
+  const setCitiesByRegionId = async (id: number) => {
+    const citiesResponse = await getCitiesByRegionId(id).then((res) => {
+      if (res.status === 200) {
```
skyboyer, line 50: what other status codes possible? I was sure that unlike raw `fetch`, `axios` will not resolve promise for cases like 404 or 500. Or may we here "positive but useless code" like 203?

skyboyer (commented):
aaaawesome refactoring; have few comments but comparing to refactoring you've done they are pretty minor

src/views/Profile/PersonalInfo/PersonalInfo.tsx
skyboyer: extra cool that we don't need to disable ESLint anymore.

just fyi, if we ever need to disable ESLint rule per-line, it's way better to disable specific rule(s):

```
// eslint-disable-next-line typescript-eslint/unsafe-assignment
```

this way we do many useful things:

1. keep catching _other_ issues
2. inform a reader what exactly was wrong - at that moment at least(since due to changes in other lines through time ESLint might stop complaining about this particular line so without a hint reader might be confused - "why this was disabled when there is no error? should I WORRY?!")

src/views/Profile/PersonalInfo/PersonalInfo.tsx
skyboyer: cool!

src/views/Profile/AuthorCreation/AuthorCreation.tsx
```diff
@@ -0,0 +1,8 @@
+import React from 'react';
+import { PersonalInfo } from '../PersonalInfo/PersonalInfo';
+
+const AuthorCreation: React.FC = () => {
```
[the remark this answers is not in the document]
skyboyer: but here you don't connect it to Redux. Or do I misunderstand? may you pleae give more context details?

Typically we create one component for editing and adding and pass different callback prop from the outside - whether it's about adding new thing or updating existing - similarly we may want to have slightly different validation, but not neccessarily

src/views/Profile/PersonalInfo/RegionCityHandler.tsx
```diff
@@ -0,0 +1,175 @@
...
+          onInputChange={(_, value) =>
+            inputRegionChangeHandler(_, value, 'regionId')
+          }
+          renderInput={(params) => (
+            <TextField
+              variant="outlined"
+              className={classes.TextField}
+              required
+              // eslint-disable-next-line
+              {...params}
```
skyboyer, line 137: typically we want `{...params}` at the very end or at the very beginning; putting it in the middle is not a technically wrong, but is not that typical and may cause a confusion if some props can be passed/overriden from outside and some cannot

src/views/Profile/PersonalInfo/RegionCityHandler.tsx
```diff
@@ -0,0 +1,175 @@
...
+          getOptionLabel={(city: ICity) => city.name}
+          getOptionSelected={(option, value) => option.id === value.id}
+          onChange={(_, value) => inputCityChangeHandler(_, value, 'cityId')}
+          renderInput={(params) => (
+            <TextField
+              variant="outlined"
+              className={classes.TextField}
+              required
+              // eslint-disable-next-line
+              {...params}
```
skyboyer, line 163: the same, `{...params}` in the middle better to be avoided

src/views/Profile/PersonalInfo/RegionCityHandler.tsx
dima-diachkov: moved to the very end

src/views/Profile/PersonalInfo/RegionCityHandler.tsx
dima-diachkov: moved to the very end

src/views/Profile/PersonalInfo/RegionCityHandler.tsx
```diff
@@ -0,0 +1,175 @@
...
+    try {
+      const res = await getRegionByCityId(id);
+      if (res.status === 200) {
+        const regionResponse = res.data;
+        setSelectRegions([regionResponse]);
+        setRegionForValue(regionResponse);
+      }
+    } catch (error) {
+      // eslint-disable-next-line no-console
+      console.error(error);
```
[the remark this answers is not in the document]
dima-diachkov: ok, so we can leave it as it is

src/views/Profile/PersonalInfo/RegionCityHandler.tsx
```diff
@@ -133,11 +133,11 @@ const RegionCityHandler = ({
               variant="outlined"
               className={classes.TextField}
               required
-              // eslint-disable-next-line
-              {...params}
               label={i18n.t(langTokens.admin.region)}
               name="region"
               onBlur={blurHandler}
+              // eslint-disable-next-line
```
skyboyer, line 139: which exact ESLint rule does it disable?

telling in advance: disabling ESLing completely, either to line or to whole file, is never a good idea
if you really need that, just list rules you are about to disable:

```
// eslint-disable-next-line rule1,rule2,rule3
```

dima-diachkov: my fault. fixed it.
unfortunately, this component accepts a lot of props and it is allowed in the rule to disable it for this reason.
https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/jsx-props-no-spreading.md

PR #806, opened 2023-02-15
add processing delete author

dima-diachkov:
Pull Request Template
 

### Type of Pull Request *

### Related links
 
**Issue link**
[#626 'Адміністрування' page: The Author record is displayed in the 'Список авторів' table after deleting](https://github.com/ita-social-projects/dokazovi-requirements/issues/626)
 
**Story link**
[#553 “Список авторів” table](https://github.com/ita-social-projects/dokazovi-requirements/issues/553)
 
### Description *
the author is deleted in the local store and a request is sent to the server for deletion but currently the server is returning 500 error

### DIFF
--- a/src/models/experts/reducers.ts
+++ b/src/models/experts/reducers.ts
@@ -71,11 +71,6 @@ export const expertsSlice = createSlice({
     ...getAsyncActionsReducer(fetchExperts as any),
     ...getAsyncActionsReducer(fetchExpertsAutorsList as any),
     ...getAsyncActionsReducer(deleteAuthor as any),
-    [deleteAuthor.fulfilled.type]: (state, action) => {
-      state.data.expertIds = state.data.expertIds.filter(
-        (id) => id !== action.payload,
-      );
-    },
   },
 });
 
--- a/src/models/experts/reducers.ts
+++ b/src/models/experts/reducers.ts
@@ -8,7 +8,11 @@ import {
   ISortAutorsList,
   ITextFields,
 } from './types';
-import { fetchExperts, fetchExpertsAutorsList } from './asyncActions';
+import {
+  deleteAuthor,
+  fetchExperts,
+  fetchExpertsAutorsList,
+} from './asyncActions';
 import { getAsyncActionsReducer } from '../helpers/asyncActions';
 
 const initialState: IExpertsState = {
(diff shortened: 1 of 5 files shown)


src/models/experts/reducers.ts
```diff
@@ -66,6 +70,12 @@ export const expertsSlice = createSlice({
   extraReducers: {
     ...getAsyncActionsReducer(fetchExperts as any),
     ...getAsyncActionsReducer(fetchExpertsAutorsList as any),
+    ...getAsyncActionsReducer(deleteAuthor as any),
+    [deleteAuthor.fulfilled.type]: (state, action) => {
+      state.data.expertIds = state.data.expertIds.filter(
```
skyboyer, line 75: I'm not sure but I feel like this way you override `[asyncAction.fulfilled.type]` from `getAsyncActionsReducer` so 

```
    state.loading = LoadingStatusEnum.succeeded;
    state[key] = action.payload;
```

is never executed. May you verify that? Sure, when deleting author it does not matter but if my assumption is correct - it may shoot us in different place(and we better to extend https://github.com/ita-social-projects/dokazovi-fe/blob/17ed7fc04b13798a0b4f26e16d3000b8c16db237/src/models/helpers/asyncActions.ts to pass custom logic into `getAsyncActionsReducer` as an argument)

src/old/lib/utilities/API/api.ts
```diff
@@ -311,3 +311,9 @@ export const updatePlatformInformation = async (
 ): Promise<AxiosResponse<PlatformInformationType>> => {
   return instance.put(`/platform-information/`, requestBody);
 };
+
+export const deleteAuthorById = async (
+  id: number,
+): Promise<AxiosResponse<PostsResponseType>> => {
```
skyboyer, line 317: now I see we have `Promise<AxiosResponse<` everywhere.

but just as information sharing, there is `AxiosPromise<` type which provides exact the same combination

src/views/Profile/AuthorsList/AuthorsDeletingDialog.tsx
```diff
@@ -22,11 +24,13 @@ const AuthorsDeletingDialog: React.FC<IAuthorsDelDialogProps> = (props) => {
   const classes = useStyles();
   const { id, fullName, isAllowedToDelete } = props;
   const [openDialog, setOpenDialog] = useState(false);
+  const [bounddeleteAuthor] = useActions([deleteAuthor]);
```
skyboyer, line 27: nit: consistent casing `boundDeleteAuthor` has better readability

skyboyer (commented):
have few small comments

src/old/lib/utilities/API/api.ts
dima-diachkov: Yes, I did as it was in the project.
I didn't know about it. indeed AxiosPromise provides the same behavior. thanks for the info)

src/models/experts/reducers.ts
dima-diachkov: You're absolutely right. I redid it in accordance with the logic of work `getAsyncActionsReducer`

skyboyer (approved):

PR #808, opened 2023-04-06
edit author

dima-diachkov:
Pull Request Template
 

### Type of Pull Request *

### Related links
 
**Issue link**
[#628 [Bug Report] 'Адміністрування' page: The 'Персональна інформація' page is not opened on the site after clicking the edit icon](https://github.com/ita-social-projects/dokazovi-requirements/issues/628 )
 
**Story link**
[#553 “Список авторів” table ]( https://github.com/ita-social-projects/dokazovi-requirements/issues/553)
 
### Description *

Added the ability to edit an existing author

image

### DIFF
--- a/src/views/Profile/PersonalInfo/PersonalInfo.tsx
+++ b/src/views/Profile/PersonalInfo/PersonalInfo.tsx
@@ -16,40 +16,58 @@ import { useStyles } from './styles/PersonalInfo.styles';
 import { ErrorField } from './ErrorField';
 import { regex } from './constants/regex';
 import { fields } from './constants/fields';
-import { IErrorFields, INewAuthorValues, IVisitFields } from './types';
+import {
+  IEditAuthorProps,
+  IErrorFields,
+  INewAuthorValues,
+  IVisitFields,
+} from './types';
 import i18n, { langTokens } from '../../../locales/localizationInit';
 import RegionCityHandler from './RegionCityHandler';
 import { validation } from './constants/validation';
 import { validateInput } from './utilities/validateInput';
 
-export const PersonalInfo: React.FC = () => {
+export const PersonalInfo: React.FC<IEditAuthorProps> = ({ author }) => {
   const classes = useStyles();
 
   const [newAuthorValues, setNewAuthorValues] = useState<INewAuthorValues>({
-    avatar: '',
-    firstName: '',
-    lastName: '',
-    regionId: 0,
-    cityId: 0,
-    bio: '',
-    email: '',
-    socialNetwork: [null, null, null, null, null],
+    avatar: author ? author.avatar : '',
+    firstName: author ? author.firstName : '',
+    lastName: author ? author.lastName : '',
+    regionId: author ? author.region.id : null,
+    cityId: author ? author.mainInstitution.city.id : null,
+    bio: author ? author.bio : '',
+    email: author ? author.email : '',
+    socialNetwork: author
+      ? author.socialNetworks
+      : [null, null, null, null, null],
+    work: author ? author.mainInstitution.name : null,
   });
   const [errorFields, setErrorFields] = useState<IErrorFields>({
-    avatar: i18n.t(langTokens.admin.pictureRequired),
-    lastName: i18n.t(langTokens.admin.fieldCantBeEmpty),
-    firstName: i18n.t(langTokens.admin.fieldCantBeEmpty),
-    regionId: i18n.t(langTokens.admin.fieldCantBeEmpty),
-    cityId: i18n.t(langTokens.admin.fieldCantBeEmpty),
-    email: i18n.t(langTokens.admin.fieldCantBeEmpty),
-    work: i18n.t(langTokens.admin.fieldCantBeEmpty),
-    bio: i18n.t(langTokens.admin.fieldCantBeEmpty),
+    avatar: author?.avatar ? '' : i18n.t(langTokens.admin.pictureRequired),
+    lastName: author?.lastName ? '' : i18n.t(langTokens.admin.fieldCantBeEmpty),
+    firstName: author?.firstName
+      ? ''
+      : i18n.t(langTokens.admin.fieldCantBeEmpty),
+    regionId: author?.region.id
+      ? ''
+      : i18n.t(langTokens.admin.fieldCantBeEmpty),
+    cityId: author?.mainInstitution.city.id
+      ? ''
+      : i18n.t(langTokens.admin.fieldCantBeEmpty),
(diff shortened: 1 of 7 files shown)


src/views/Profile/PersonalInfo/RegionCityHandler.tsx
```diff
@@ -151,6 +155,11 @@ const RegionCityHandler = ({
           disablePortal
           options={selectCities || cities}
           fullWidth
+          value={
+            newAuthorValues.cityId
+              ? cities.find((city) => city.id === newAuthorValues.cityId)
```
skyboyer, line 160: I think ternary is not really needed here:

```
value={cities.find((city) => city.id === newAuthorValues.cityId)}
```

will return `undefined` if `newAuthroValues.cityId` is `null`, `undefined` or empty string

src/views/Profile/PersonalInfo/RegionCityHandler.tsx
```diff
@@ -122,7 +122,11 @@ const RegionCityHandler = ({
           disablePortal
           options={selectRegions || regions}
           fullWidth
-          value={regionForValue}
+          value={
+            newAuthorValues.regionId
+              ? regions.find((region) => region.id === newAuthorValues.regionId)
```
skyboyer, line 127: why we need `regionForValue` when selection does not match any region or is empty?

src/views/Profile/PersonalInfo/PersonalInfo.tsx
```diff
@@ -16,40 +16,58 @@ import { useStyles } from './styles/PersonalInfo.styles';
...
   const [errorFields, setErrorFields] = useState<IErrorFields>({
-    avatar: i18n.t(langTokens.admin.pictureRequired),
-    lastName: i18n.t(langTokens.admin.fieldCantBeEmpty),
-    firstName: i18n.t(langTokens.admin.fieldCantBeEmpty),
-    regionId: i18n.t(langTokens.admin.fieldCantBeEmpty),
-    cityId: i18n.t(langTokens.admin.fieldCantBeEmpty),
-    email: i18n.t(langTokens.admin.fieldCantBeEmpty),
-    work: i18n.t(langTokens.admin.fieldCantBeEmpty),
-    bio: i18n.t(langTokens.admin.fieldCantBeEmpty),
+    avatar: author?.avatar ? '' : i18n.t(langTokens.admin.pictureRequired),
```
skyboyer, line 47: but why do we need to keep error messages in the state? what if we will calculate it dynamically with `useMemo`?

```
const errorMessages = useMemo(() => {
  const errors = {
    email: '',
    lastName: '',
    // ...
  };
  if (!author?.email) {
    errors[email] =  i18n.t(langTokens.admin.fieldCantBeEmpty),;
  } else if (value && !regex.validEmail.test(author.email)) {
      errors.email = i18n.t(langTokens.admin.wrongEmail);
  }
  .....
}, [author?.email, author?.bio, author?.region.id])
```

then we would not need to call `setErrorFields` just to update error messages. And keeping all the validation logic in single place may be beneficial - though, this block might grow in size, it's a typical case where for validation we need to check _several_ fields which depend on each other. E.g. just an synthetic example - for address block which is optional, we either keep all the fields empty, but if we provide street name, we must also provide a city ID. When validation in single place, it's easier to implement checks like that.

src/views/Profile/PersonalInfo/RegionCityHandler.tsx
dima-diachkov: maybe we don't need it at all. I removed it and used `newAuthorValues` to store the value.

src/views/Profile/PersonalInfo/PersonalInfo.tsx
dima-diachkov: I applied. If I understand correctly, this significantly reduced and improved the code.

src/views/Profile/PersonalInfo/PersonalInfo.tsx
```diff
@@ -41,34 +41,94 @@ export const PersonalInfo: React.FC<IEditAuthorProps> = ({ author }) => {
...
-      ? ''
-      : i18n.t(langTokens.admin.fieldCantBeEmpty),
-    regionId: author?.region.id
-      ? ''
-      : i18n.t(langTokens.admin.fieldCantBeEmpty),
-    cityId: author?.mainInstitution.city.id
+
+  const errorMessages = useMemo(() => {
+    const errors: IErrorFields = {
+      avatar: author?.avatar ? '' : i18n.t(langTokens.admin.pictureRequired),
```
skyboyer, line 49: 1. here we are checking fields from `author`, later we check fields from `newAuthorValues`, please explain, I'm confused why we are validating the same fields from props and from state
2. even if we need to validate from both objects, we should add `author` to dependency array of `useMemo`. I'll raise this, we have `react-hooks/exhaustive-deps` disabled in our ESLint config and that's pretty dangerous.
3. whether or not we want to validate both `author` and `newAuthorValues` I believe it's better to change pattern for validation. Current:
```
errors = {
  email: ...
  ....
};
error.email = validateInput(...)
```
just overwrites messages from first block with result of `validateInput` call. It might be not a defect if you get desired error message finally, but it's really, really confusing.
So rule of thumb: setting single error key just once:
```
const error = {
...
  email: '',
...
};
if (...) {
  error.email = ...
} else if {
  error.email = ....
}
```

src/views/Profile/PersonalInfo/PersonalInfo.tsx
```diff
@@ -220,13 +210,7 @@ export const PersonalInfo: React.FC<IEditAuthorProps> = ({ author }) => {
...
   useEffect(() => {
     toggleButtonHandler();
-  }, [errorFields]);
-
-  useEffect(() => {
-    if (newAuthorValues.socialNetwork.some((network) => network != null)) {
-      setErrorFields({ ...errorFields, socialNetwoks: '' });
-    }
-  }, [newAuthorValues.socialNetwork]);
+  }, [errorMessages]);
```
skyboyer, line 213: `toggleButton`(I would name it `isSaveDisabled` instead) does not need to be a state variable:

```
const isSaveDisabled = useMemo(
  () => Object.values(errorMessages).some(message => message),
  [errorMessages]
);
```

skyboyer (commented):
really like new changes. however have few comments on them.

skyboyer (approved):
I'm not sure if all the commits have been pushed yet(regarding minor things we have had conversations), but so far looks good and I agree with merging

PR #811, opened 2023-08-24
feat(author): add filter publications by status

OlehFilyak:
Pull Request Template
 

### Type of Pull Request *

### Related links
 
**Issue link**
[#603 (Author) Filter publications by Status]( )](https://github.com/ita-social-projects/dokazovi-requirements/issues/603)
 
**Story link**
[#603 Story]( https://github.com/ita-social-projects/dokazovi-requirements/issues/603#issue-1344388099)
 
### Description *
I added the filter by status to the author filter and did not change the order of filters in the admin filters. That's why I didn't remove this filter from the isAdmin condition.
 
"*" - mandatory field

### DIFF
--- a/src/views/Profile/AdminTable/AdminTableFilters.tsx
+++ b/src/views/Profile/AdminTable/AdminTableFilters.tsx
@@ -93,6 +93,16 @@ const AdminTableFilters: React.FC = () => {
           filter={QueryTypeEnum.POST_TYPES}
         />
       </Grid>
+      {!isAdmin && (
+        <Grid item>
+          <AdminFilter
+            setChanges={boundedSetFilter}
+            allOptions={allPostStatuses}
+            selected={filters.statuses}
+            filter={QueryTypeEnum.STATUSES}
+          />
+        </Grid>
+      )}
       {isAdmin && (
         <>
           <Grid item md={2} className={classes.textField}>
--- a/src/views/Profile/AdminTable/AdminTableFilters.tsx
+++ b/src/views/Profile/AdminTable/AdminTableFilters.tsx
@@ -66,16 +66,16 @@ const AdminTableFilters: React.FC = () => {
           filter={QueryTypeEnum.DIRECTIONS}
         />
       </Grid>
+      <Grid item>
+        <AdminFilter
+          setChanges={boundedSetFilter}
+          allOptions={allPostStatuses}
+          selected={filters.statuses}
+          filter={QueryTypeEnum.STATUSES}
+        />
+      </Grid>
       {isAdmin && (
         <>
-          <Grid item>
-            <AdminFilter
-              setChanges={boundedSetFilter}
-              allOptions={allPostStatuses}
-              selected={filters.statuses}
-              filter={QueryTypeEnum.STATUSES}
-            />
-          </Grid>
           <Grid item md={3}>
             <AdminDatePicker
               start={date.start}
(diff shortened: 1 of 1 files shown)


src/views/Profile/AdminTable/AdminTableFilters.tsx
```diff
@@ -93,6 +93,16 @@ const AdminTableFilters: React.FC = () => {
           filter={QueryTypeEnum.POST_TYPES}
         />
       </Grid>
+      {!isAdmin && (
```
skyboyer: so I see filter by status is exactly the same to admin and to normal user, but appears in different ordering among other filters. May we want to have consistency here? Then instead of having

```
            <AdminFilter
              setChanges={boundedSetFilter}
              allOptions={allPostStatuses}
              selected={filters.statuses}
              filter={QueryTypeEnum.STATUSES}
            />
```
twice, we could have it just once.

Reasoning:

1. code duplication means we can introduce an error or different behavior by accident
2. consistent UI, I believe, is better - although admin user barely logs in as normal user on Tuesdays 😅 so probably nobody can notice this difference

skyboyer (approved):
overall changes are good, but if it's possible to place filter consistently _and_ to avoid code duplication - it would be really nice!

src/views/Profile/AdminTable/AdminTableFilters.tsx
OlehFilyak: Ok, I understand. Thank you for your answer.
I think we can resolve it in two different ways:
1. If we change the order of components in the author's panel. Is it OK if there is a next order: 'теми статуси типи' instead of 'теми типи статуси'.
2. We can change the order with CSS. But this will greatly complicate the code understanding for other developers.
What should I do?

OlehFilyak: > so I see filter by status is exactly the same to admin and to normal user, but appears in different ordering among other filters. May we want to have consistency here? Then instead of having
> 
> ```
>             <AdminFilter
>               setChanges={boundedSetFilter}
>               allOptions={allPostStatuses}
>               selected={filters.statuses}
>               filter={QueryTypeEnum.STATUSES}
>             />
> ```
> 
> twice, we could have it just once.
> 
> Reasoning:
> 
> 1. code duplication means we can introduce an error or different behavior by accident
> 2. consistent UI, I believe, is better - although admin user barely logs in as normal user on Tuesdays 😅 so probably nobody can notice this difference

Ok, I understand. Thank you for your answer.
I think we can resolve it in two different ways:

If we change the order of components in the author's panel. Is it OK if there is a next order: 'теми статуси типи' instead of 'теми типи статуси'.
We can change the order with CSS. But this will greatly complicate the code understanding for other developers.
What should I do?

OlehFilyak: I committed to the first solution. Check it, please. I also tested admin filters there nothing changed.

skyboyer: look, I'm a technical expert and have no authority to make decisions regarding UI or requirements :) This is just my assumption that this change would be for better. Plz communicate to BA/PM and confirm this change with them.

Regarding way to implement, yes, reordering in the code way better for understanding. So I agree, approach 1 is thing to take.

OlehFilyak: > 

I contacted BA and he approved the changes to the filter sequence. Please check the implementation of the first approach.

PR #812, opened 2023-08-31
Develop release #812 fix conflicts

dima-diachkov:
resolved conflicts before pulling changes from develop into release

### Type of Pull Request *
- [x] NONE (if none of the other choices apply. For example, tooling, build system, CI, docs, etc.)


src/old/lib/utilities/API/api.ts
```diff
@@ -166,6 +166,9 @@ export const setFakePostViewsCounter = async (
       views,
     },
   });
+  // const result = instance.post(`/post/set-fake-view/${id}?views=${views}`);
```
skyboyer, line 169: is it needed? :)

skyboyer (approved):

src/old/lib/utilities/API/api.ts
dima-diachkov: Most likely no. In any case, unfortunately, there is quite a lot of this. it probably even needs to be highlighted in a separate issue, but now the task is to upload the changes from develop to release :)

PR #814, opened 2023-09-04
Issue 679 to redesign filter button by date

OlehFilyak:
Pull Request Template
 

### Type of Pull Request *
- [X] FEATURE (non-breaking change which adds functionality)

### Related links
 
**Issue link**
[#679 To redesign filter button by date ](https://github.com/ita-social-projects/dokazovi-requirements/issues/679)
 )
 
**Story link**
[#679 1 Story]( https://github.com/ita-social-projects/dokazovi-requirements/issues/679#issue-1871973783)
 
### Description *
At first, when you don't choose any date, there are no styles for today's day. When you choose someday styles add. If you choose the same day we will reset the styles and the value in the global state.


src/views/Profile/AdminTable/AdminDatePicker.tsx
```diff
@@ -32,18 +35,100 @@ interface IMaterialsDate {
...
+  bgColor: string;
+  value: string;
+}
+
+const datePickerReducer = (
+  state: {
+    firstSelectedDay: ISelectedDay;
+    secondSelectedDay: ISelectedDay;
+  },
+  action: { type: string; payload: ISelectedDay },
```
skyboyer: it always makes sense to type `type` part. then TS will save our backs when we make a typo in `type` value

src/views/Profile/AdminTable/AdminDatePicker.tsx
```diff
@@ -32,18 +35,100 @@ interface IMaterialsDate {
...
+): {
+  firstSelectedDay: ISelectedDay;
+  secondSelectedDay: ISelectedDay;
+} => {
+  switch (action.type) {
+    case 'SET_SELECTED_DAY':
```
skyboyer: instead of using generic `SET_SELECTED_DAY` type, I propose to have separate `SET_START_DAY` and `SET_END_DAY` right from the start. This give us 2 benefits: 
1. better readability
a. (no `if .. ` nested into `switch/case` - the less nesting the easier to understand) 
b. explicit types for `START` and `END` instead of passing this as a part of payload
2, more flexibility: maybe it has not been stated yet, but typically it's expected that when we set end date before current start date or in opposite - setting start date after current end date, we either clear or update counterpart(e.g. when setting start date before current end date we either clear end date at all or update it to be equal to start date); even if this was not required yet, extra logic is still easier to apply with separate action types

src/views/Profile/AdminTable/AdminDatePicker.tsx
```diff
@@ -32,18 +35,100 @@ interface IMaterialsDate {
...
+  return state;
+};
+
 export const AdminDatePicker: React.FC<IMaterialsDate> = ({
   start,
   end,
   setChanges,
 }) => {
+  const [datePickerState, dispatch] = useReducer(datePickerReducer, {
+    firstSelectedDay: { position: 'start', bgColor: 'transparent', value: '' },
```
skyboyer: am I right that `firstSelectedDay` is always `position: 'start'` when `secondSelectedDay` is always `position: 'end'`?

src/views/Profile/AdminTable/AdminDatePicker.tsx
```diff
@@ -32,18 +35,100 @@ interface IMaterialsDate {
...
-    setChanges({
-      option,
-      date: requestDate(value),
-    });
+  const today = new Date();
+
+  const handleDatePicker = (option: string, value: Date | null) => {
+    if (value) {
+      const changedValue = value.setHours(23, 59, 59, 0).toString();
+      if (
```
skyboyer: I'm a bit lost here. Could you please add some short description why do we check what we check here?

src/views/Profile/AdminTable/AdminDatePicker.tsx
```diff
@@ -54,39 +139,49 @@ export const AdminDatePicker: React.FC<IMaterialsDate> = ({
...
-              {t(langTokens.admin.to)} {parsDate(end)}
-            </div>
-          )}
+          renderValue={() =>
+            !datePickerState.firstSelectedDay.value &&
+            !datePickerState.secondSelectedDay.value ? (
+              <DateRangeOutlinedIcon />
+            ) : (
+              <div>
+                {t(langTokens.admin.from)} {parsDate(start)}{' '}
```
skyboyer: I strongly suggest to code this as a single translation like

```
З {{parsedStartDate}} по {{parsedEndDate}}
```

Something like https://github.com/ita-social-projects/dokazovi-fe/blob/develop/src/locales/uk/parts/admin.ts#L79

You see, in different languages structure may be different, e.g. "from" and "to" might be combined into single word between dates or maybe grammar rules requires us to add text _after_ dates, not between. Having this coded in single solid message template gives much more flexibility with no disadvantages.

src/views/Profile/AdminTable/AdminDatePicker.tsx
OlehFilyak: Yes, you are right. I'll change it according to what you wrote above.

src/views/Profile/AdminTable/AdminDatePicker.tsx
OlehFilyak: Okay. I'll try. Value can be Date, or null. In this line:
                onChange={(value) => {
                  handleDatePicker('start', value);
                }}
                We had an error
                Argument of type 'MaterialUiPickersDate' is not assignable to parameter of type 'Date'.
  Type 'null' is not assignable to type 'Date'.

skyboyer: well, this explains `if (value)` part but not `option ==='end' && changedValue != ....`

also I'm not in the context why `setChanges` gets just a single value. does the parent element also have "start + end dates"? does it make sense to rely just on the state in parent instead of having 2 different sources of the truth?

src/locales/uk/parts/admin.ts
```diff
@@ -46,8 +46,7 @@ export const admin = {
   changeDateSuccess: 'Дату публікації змінено',
   scheduleDateSuccess: 'Дату публікації заплановано',
   schedulePublishTitle: 'Запланувати публікацію?',
-  from: 'з',
```
skyboyer, line 49: it's not used anywhere else, is it?

src/views/Profile/AdminTable/AdminDatePicker.tsx
```diff
@@ -32,18 +35,93 @@ interface IMaterialsDate {
...
+  bgColor: 'transparent' | 'grey';
+  value: string | '';
+};
+
+const datePickerReducer = (
+  state: {
+    firstSelectedDay: ISelectedDayType;
+    secondSelectedDay: ISelectedDayType;
+  },
+  action: { type: string; payload: ISelectedDayType },
```
skyboyer: `type` is better to be typed as [a discriminated union](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#discriminated-unions)

src/views/Profile/AdminTable/AdminDatePicker.tsx
```diff
@@ -54,39 +132,51 @@ export const AdminDatePicker: React.FC<IMaterialsDate> = ({
...
+                autoOk
+                orientation="portrait"
+                variant="static"
+                openTo="date"
+                value={start}
+                onChange={(value) => {
+                  handleDatePicker('start', value);
+                }}
+                format="dd-mm-yyyy"
+                maxDate={today}
```
skyboyer: how does code behave if we select `start` > `end`?

src/views/Profile/AdminTable/AdminDatePicker.tsx
```diff
@@ -54,39 +132,51 @@ export const AdminDatePicker: React.FC<IMaterialsDate> = ({
...
-              variant="static"
-              openTo="date"
-              value={end}
-              onChange={(value) => {
-                handleDatePicker('end', value);
-              }}
-              format="dd-mm-yyyy"
-              minDate={start}
-            />
+            <Box className={datepickerClassesStart.datePicker}>
```
skyboyer, line 158: I think, instead of putting color into a reducer and then create 2 different `datepickerClassesStart` and `datepickerClassesEnd` we could instead:

```
  const datepickerClassesSelectedDate = useStylesForDatePicker({
    bgColorForDatePicker: 'grey',
  });
  const datepickerClassesNoSelection = useStylesForDatePicker({
    bgColorForDatePicker: 'transparent',
  });
......
<Box className={start ? datepickerClassesSelectedDate.datePicker :  datepickerClassesNoSelection.datePicker }>
....
<Box className={end ? datepickerClassesSelectedDate.datePicker :  datepickerClassesNoSelection.datePicker }>
```

A bit repetitive? may be. but we don't have any other difference between Start Date and End Date selector, we just need to suppress current day highlighted, right?

Also this would allow us to get rid of extra state(`useReducer` it's also a state) and consume `end`/`start` again.

--------

Ideally, I'd like to have `DokazoviDatePicker.tsx` which would implement logic under the hood:

1. no highlighting for the current day if no selection
2. clicking currently selected day unselects it

This way we would re-use it across the app getting unified UX.

Do we use DatePicker elsewhere?

src/views/Profile/AdminTable/AdminDatePicker.tsx
```diff
@@ -32,18 +35,93 @@ interface IMaterialsDate {
...
-  const handleDatePicker = (option: string, value) => {
-    setChanges({
-      option,
-      date: requestDate(value),
-    });
+  const today = new Date();
+
+  const handleDatePicker = (option: string, value: Date | null) => {
+    if (value) {
+      const changedValue = value.setHours(23, 59, 59, 0).toString();
```
skyboyer: just to understand better: why to set to `23:59:59` not to `00:00:00`?

src/views/Profile/AdminTable/styles/AdminDatePicker.styles.ts
```diff
@@ -13,3 +13,12 @@ export const useStyles = makeStyles((theme: Theme) => ({
     },
   },
 }));
+
+export const useStylesForDatePicker = makeStyles({
+  datePicker: {
+    '& .MuiPickersDay-daySelected': {
```
skyboyer, line 19: btw btw, I see `disableHighlightToday` for [docs on](https://mui.com/x/api/date-pickers/date-picker/#DatePicker-prop-disableHighlightToday) newer version of material-ui. And looking for this property over legacy 3.3.10 I see [some usage](https://github.com/search?q=repo%3Amui%2Fmaterial-ui-pickers%20disableHighlightToday&type=code). May it be that it worked also in 3.3.10?

skyboyer: I've checked in codesandbox and surprisingly it does not allow me to add `@material-ui/pickers@3.3.10` like it if was unsupported. So I'm asking you to check this on Dokazovi,

src/locales/uk/parts/admin.ts
OlehFilyak: No, this code isn't used anywhere. I searched in the project before I removed this line.

src/views/Profile/AdminTable/AdminDatePicker.tsx
OlehFilyak: Sorry, I didn't understand you before, but now I changed it.

src/views/Profile/AdminTable/AdminDatePicker.tsx
OlehFilyak: We have one handleDatePicker and when we send two different flags - 'start', and 'end'. When we click on the first DatePicker we automatically send 'start', and when we click on the second DatePicker we automatically send 'end'. This logic was written before I started rewriting this component.

src/views/Profile/AdminTable/AdminDatePicker.tsx
OlehFilyak: We use DatePicker in SetPublicationDateModal. But not use AdminDatePicker.

src/views/Profile/AdminTable/AdminDatePicker.tsx
OlehFilyak: We need to set 'setHours' for the correct comparison without different hours. If we set 00:00:00 it'll automatically transfer us to the previous day and apply logic for it.

src/views/Profile/AdminTable/AdminDatePicker.tsx
skyboyer: let's me explain what I mean

say, today is 06/09

i'm setting end date through second picker to 31/08
now I'm setting start date through first picker to 02/09

then start date> end date

how will it behave in this case?

src/views/Profile/AdminTable/AdminDatePicker.tsx
skyboyer: do we expect current day to be highlighted in SetPublicationDateModal?

src/views/Profile/AdminTable/AdminDatePicker.tsx
skyboyer: sorry, I don't get it :) which code exactly threats "2023-09-06 00:00:00" as September, 5th?

src/views/Profile/AdminTable/styles/AdminDatePicker.styles.ts
OlehFilyak: I've already checked it, and it doesn't allow me to add it too.

src/views/Profile/AdminTable/AdminDatePicker.tsx
OlehFilyak: If we write value.setHours(0, 0, 0, 0).toString() and choose 2023-09-06, we take hilghlighting for 2023-09-05
Знімок екрана 2023-09-06 о 15 43 41

src/views/Profile/AdminTable/styles/AdminDatePicker.styles.ts
OlehFilyak: We can't add "disableHighlightToday", because we don't have this property in our library version. I find it in node modules.

src/views/Profile/AdminTable/AdminDatePicker.tsx
OlehFilyak: No, I asked Denis and he said that we need there the same behavior as this.

src/views/Profile/AdminTable/AdminDatePicker.tsx
skyboyer: I see. I suspect depending on current time zone, setting date to 23:59:59 may instead highlight next day
May you check with changing local timezone e.g. to USA time?

src/views/Profile/AdminTable/AdminDatePicker.tsx
skyboyer: understood. Then I propose to extract this into a separate component - also `<DatePicker` but which will have all this extra styling. All extra props should be passed down to `material-ui/pickers`' DatePicker. Then replacing it will just a change in `import .. from ` line

Easier to reuse, no risks of inconsistent changes

src/views/Profile/AdminTable/styles/AdminDatePicker.styles.ts
skyboyer: > I find it in node modules.

I guess, "I did not find", right?

src/views/Profile/AdminTable/AdminDatePicker.tsx
skyboyer: so will you extract this into a separate component, so reducer here will not orchestrate colors?

OlehFilyak: Sorry, I didn't write it. I tried to do that, but if we move in this direction, we greatly increase the duplication and confusion of the code, because style management is closely related to value in this component. I tried to rewrite the component without extracting this logic. It's a specific component, and I don't think we should extract it in a separate file. You can look at my attempt to rewrite the component in a zip file.
[AdminDatePicker 2.tsx.zip](https://github.com/ita-social-projects/dokazovi-fe/files/12584286/AdminDatePicker.2.tsx.zip)

src/views/Profile/AdminTable/AdminDatePicker.tsx
skyboyer: so how does it work in such a case?

OlehFilyak: I changed the logic in this component. And when you set the end date through the second picker to 31/08, now you can't set the start date through the first picker to 02/09. The start day must be earlier than the end day.

src/views/Profile/AdminTable/AdminDatePicker.tsx
skyboyer: this is what I thought about:

```
import { DatePicker as MuiDatePicker } from '@material-ui/pickers';
import {createMuiTheme, makeStyles} from '@material-ui/core';

const useStyles = makeStyles((theme) => ({
  container: {
    display: 'contents',
    '& .MuiPickersDay-daySelected': {
      backgroundColor: ({ isHighlightingEnabled }) => isHighlightingEnabled ? theme.palette.primary.main : 'transparent',
      color: ({ isHighlightingEnabled }) => isHighlightingEnabled ? theme.palette.primary.contrastText : theme.palette.text.primary
    }
  }
}))

export function DatePicker(props) {
  const isHighlightingEnabled = !!props.value;
  const classes = useStyles({ isHighlightingEnabled })
  const onChange = useCallback((value) => {
    if (String(value) === String(props.value)) {
      props.onChange(null)
    } else {
      props.onChange(value)
    }
  }, [props.value, props.onChange])

  return (<div className={classes.container}>
    <MuiDatePicker 
      {...props} 
      variant="static"
      onChange={onChange} 
    />
  </div>)
}
```

and then instead of `import { DatePicker } from '@material-ui/pickers'` we could `import { DatePicker } from '../ourCustomDatePicker.tsx'` and use it just like if `material-ui`'s version would patched and worked with "unsetting" date and suppressing highlighting today day if `value` is `null`

the big disadvantage of that method - styles will work fine only if `variant="static"`, otherwise Dialog is rendered outside of our "extra div with custom class" and override does not work

What do you say?

PR #817, opened 2023-09-11
bugFix #652 and feat(AdminFilter) hide menu when and admin start scro…

OlehFilyak:
Pull Request Template
 

### Type of Pull Request *

### Related links
 
**Issue link**
[bugFix [#652](https://github.com/ita-social-projects/dokazovi-requirements/issues/652)]
 
### Description *
Fix the bug by adding a new feature - when the admin starts scrolling down we hide the the menu

### DIFF
--- a/src/views/Profile/AdminTable/AdminFilter.tsx
+++ b/src/views/Profile/AdminTable/AdminFilter.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable */
-import React, { useState, useEffect, useRef } from 'react';
+import React, { useState, useEffect, useCallback } from 'react';
 import MenuItem from '@material-ui/core/MenuItem';
 import FormControl from '@material-ui/core/FormControl';
 import ListItemText from '@material-ui/core/ListItemText';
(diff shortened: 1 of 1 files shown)


src/views/Profile/AdminTable/AdminFilter.tsx
```diff
@@ -51,6 +50,30 @@ export const AdminFilter: React.FC<IMaterialsFilter> = ({
 
   const classes = useStyles();
   const { t } = useTranslation();
+  const [open, setOpen] = useState(false);
+  const selectRef = useRef(null);
```
skyboyer: why do we need this ref? I don't see our code to consume it

src/views/Profile/AdminTable/AdminFilter.tsx
```diff
@@ -51,6 +50,30 @@ export const AdminFilter: React.FC<IMaterialsFilter> = ({
...
+    };
+
+    window.addEventListener('scroll', handleScroll);
+
+    return () => {
+      window.removeEventListener('scroll', handleScroll);
+    };
+  }, [open]);
+
+  const handleOpen = () => {
```
skyboyer: if you wrap `handleOpen` and `handleClose` into `useCallback()` you could use `handleClose` in the `useEffect` above without risk to fall into infinite rerender

src/views/Profile/AdminTable/AdminFilter.tsx
```diff
@@ -51,6 +50,30 @@ export const AdminFilter: React.FC<IMaterialsFilter> = ({
 
   const classes = useStyles();
   const { t } = useTranslation();
+  const [open, setOpen] = useState(false);
+  const selectRef = useRef(null);
+
+  useEffect(() => {
+    const handleScroll = () => {
+      if (open) {
```
skyboyer: but do we really need this check? I guess, calling `setOpen(false)` will work fine even if dropdown is already closed. No?

src/views/Profile/AdminTable/AdminFilter.tsx
```diff
@@ -51,29 +51,18 @@ export const AdminFilter: React.FC<IMaterialsFilter> = ({
...
   const [open, setOpen] = useState(false);
-  const selectRef = useRef(null);
 
   useEffect(() => {
-    const handleScroll = () => {
-      if (open) {
-        setOpen(false);
-      }
-    };
+    const handleScroll = () => handleClose();
```
skyboyer, line 56: you could directly use `handleClose` but using `handleScroll` as a wrapper around it is also fine

in my opinion

skyboyer (approved):

PR #820, opened 2023-09-12
bugFix #595 Fixed incorrect order of filter items on the "Матеріали" page after selecting the "Теми" filter.

SolomiiaLomnytska22:
### Type of Pull Request *

### Related links
 
**Issue link**
[bugFix #595]
 
 
### Description *
Fixed incorrect order of items in the "Теми" filter in the author profile on the "Матеріали" page.

### DIFF
--- a/src/views/Profile/AdminTable/AdminFilter.tsx
+++ b/src/views/Profile/AdminTable/AdminFilter.tsx
@@ -49,6 +49,12 @@ export const AdminFilter: React.FC<IMaterialsFilter> = ({
 }) => {
   const ALL_OPTIONS_IDS = allOptions.map((e) => e.id);
 
+  const sortedOptions = [...allOptions].sort((a, b) => {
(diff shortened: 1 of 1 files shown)


skyboyer (approved):

src/views/Profile/AdminTable/AdminFilter.tsx
```diff
@@ -49,6 +49,12 @@ export const AdminFilter: React.FC<IMaterialsFilter> = ({
 }) => {
   const ALL_OPTIONS_IDS = allOptions.map((e) => e.id);
 
+  const sortedOptions = [...allOptions].sort((a, b) => {
```
skyboyer, line 52: btw how long `allOptions` can be? may it make sense to wrap sorting into `useMemo` to avoid re-sorting on each re-render? or is list pretty small so it is not worth it?

PR #823, opened 2023-09-18
Bugfix628 dispatch new author

dima-diachkov:
Added logic for sending a new or edited author to the backend
 

### Type of Pull Request *

### Related links
 
[#628 Bug Report](https://github.com/ita-social-projects/dokazovi-requirements/issues/628 )
[#627 Bug Report](https://github.com/ita-social-projects/dokazovi-requirements/issues/627 )


src/old/lib/utilities/API/types.ts
```diff
@@ -316,3 +316,15 @@ export type GetChangeLogType = ChangeLogType & {
   title?: string;
   totalPages: number;
 };
+
+export type CreateUpdateAuthorRequestType = {
```
skyboyer, line 320: not a big deal, but I'd have 2 different types(like separate names) right from the start. 
Why it might be helpful - well, not here specifically, but in general case - 

1. when it's a type of argument we have better visibility what's passed and for what
2. again, not for this case, but it's quite often to have some fields can be set only once, per creation; so having CreateRequestType and UpdateRequestType right from the start means less headache in this scenario
3. while they are equal, sure, no need to completely duplicate whole the type definition; just using type aliases would be enough by now

src/views/Profile/PersonalInfo/PersonalInfo.tsx
```diff
@@ -29,22 +31,22 @@ import { validateInput } from './utilities/validateInput';
...
+  const history = useHistory();
 
   const [visitFields, setVisitFields] = useState<IVisitFields>({
     lastName: false,
     firstName: false,
     region: false,
     city: false,
     email: false,
-    work: false,
+    mainWorkingPlace: false,
```
skyboyer, line 42: just curious, backend has been updated but UI was not?

src/views/Profile/PersonalInfo/PersonalInfo.tsx
```diff
@@ -182,8 +184,13 @@ export const PersonalInfo: React.FC<IEditAuthorProps> = ({ author }) => {
     setVisitFields({ ...visitFields, [inputIdentifier]: true });
   };
 
-  const buttonClickHandler = () => {
-    // TODO dispatch new author to server
+  const buttonClickHandler = async (event: { preventDefault: () => void }) => {
```
skyboyer, line 187: I propose to use `MouseEvent` interface instead

skyboyer (commented):
heh, so we did not have this before. Nicely done. Have just few comments

PR #831, opened 2023-12-26
Issue634 create user account for author

### LINKED ISSUE #634: Admin: Create user account for the author
As an Admin, I want to create user account for the author

Preconditions:
Registered User with Admin permission is logged in
Choose the ‘Aдміністрування’ options from the drop-down list at the Header
Choose the “Список авторів” page 
Choose the edit option in the “Список авторів” table

Acceptance criteria:
1. Add a new button under the personal information Активувати користувача/Деактивувати користувача, depending on “Активний користувач” status. Button "Активувати користувача" should be displayed in green color, "Деактивувати користувача" in red color.
2. When the button "Активувати користувача" is pressed, should be 2 scenarios: 
2а.  Scenario 1. If user account already exists but was inactive, should be displayed text: "Акаунт користувача вже існує, ви хочете активувати акаунт чи створити новий?" and 2 buttons: "Активувати існуючий акаунт" & "Створити новий акаунт".
If the admin chooses "Активувати існуючий акаунт", author user account should be activated and a notification email letter should be sent to the email with a link to logging in.
 2b. If the admin chooses a "Створити новий акаунт" button in scenario 1, or user account didn't exist, should be dispayed a new field instead of the button for the email with the button “відправити”. After pressing the button should be sent a verification email letter with the link. 
3. When the button "Деактивувати користувача" is pressed user should lose the availability to log in as an author.
4. Email validation - email should have local-part, @ symbol, and domain name.
5. Email for the account and public email must be different, email should be checked for duplicates with the other user accounts. Show email criteria to the user. 
6. If the public email and email for the user account match, an error message should be displayed: "Пошта для акаунту має відрізнятися від публічної пошти"
7. If the email for the user account duplicates with the other user account email, an error message should be displayed: "За чією поштою вже зареєстрований користувач"
8. A link should be unique for each letter and be active for 24 hours.
9. The author should become active only after email validation and creating the password. 
10. After creating a user account, a notification email letter should be sent to the email with a link to logging in.
11. If the email is already sent, but not validated, and the link is still active, after pressing the "Активувати користувача" button, should be displayed an error message: "Лист для активації висланий, підтвердіть пошту або зачекайте n часу і спробуйте ще раз". Where n - time left for the link activation in minutes.

YuriyHryshchenko:
### Type of Pull Request *

### Related links
 
**Issue link**
[#634 Create user account for the author](https://github.com/ita-social-projects/dokazovi-requirements/issues/634 )
  
### Description *
Add ability to create user account. If author hasn't account - he has status "NEW" and he need to send email to get back link to confirm password - this functionality after confirmation adds ability to the author to login by email and password in his acc. If author had already made his acc -  he has status "ACTIVE". Admin can change status to - "DELETED" - this functionality deactivates author acc and he can`t login any more. If author has status "DELETED" admin can activate this author by changing his status to "ACTIVE" or create him a new acc like it was with status "NEW"


src/components/Form/Buttons/styles/UserAccountButton.style.tsx
```diff
@@ -0,0 +1,41 @@
...
+      },
+    },
+    createButton: {
+      backgroundColor: '#73ddff',
+      '&:hover': {
+        backgroundColor: theme.palette.info.main,
+      },
+    },
+    deactivateButton: {
+      backgroundColor: theme.palette.error.main,
```
skyboyer: maintainability-wise: I think this entry should be `theme.palette.danger.main` instead of `.error.`. Coloring active controls like a button with a color of "error state" sounds confusing. Yes, it may be exactly the same value under the hood.

src/components/Form/Buttons/styles/UserAccountButton.style.tsx
```diff
@@ -0,0 +1,41 @@
...
+      },
+    },
+    activateButton: {
+      backgroundColor: theme.palette.success.main,
+      '&:hover': {
+        backgroundColor: theme.palette.success.light,
+      },
+    },
+    createButton: {
+      backgroundColor: '#73ddff',
```
skyboyer: no existing entry in the theme palette? time to create new one!

PS I don't have a proposal for semantically meaningful name

PPS white text color over #73ddff background has terrible accessibility :(

src/components/Form/Buttons/styles/UserAccountButton.style.tsx
```diff
@@ -0,0 +1,41 @@
...
+      '&:hover': {
+        backgroundColor: theme.palette.info.main,
```
skyboyer: if `#73ddff` goes to new entry, then this value should also go from another new entry too

src/components/Form/Buttons/styles/UserAccountButton.style.tsx
```diff
@@ -0,0 +1,41 @@
+import { makeStyles, Theme } from '@material-ui/core';
+
+export const useStyles = makeStyles(
+  (theme: Theme) => ({
+    userAccountButton: {
+      flex: '0 1 300px',
+      padding: theme.spacing(3.2, 8.8),
```
skyboyer, line 7: I see padding is the same as for `BasicButton`. Would it be possible to extend(oookay, "extend" because it's not really the inheritance :) ) `<BasicButton>` instead? Something like 

```
function UserAccountButton(...) {
...
  return <BasicButton ...
````
?

I don't know what's different except the colors and `300px` starting width, so maybe re-using `BaseButton` cannot be done here. But typically we rather want our controls to have consistency in their look. And hard-coding layout-related styles(font-size, padding, starting width etc) across same-type component creates risks to that consistency.

File: src/components/Form/Buttons/UserAccountButton.tsx [hunk 5]
```diff
@@ -0,0 +1,43 @@
...
+import { useStyles } from './styles/UserAccountButton.style';
+
+type UserAccountButtonPropsType = {
+  type: 'activate' | 'deactivate' | 'create';
+  label: string;
+  action?: 'submit';
+  disabled?: boolean;
+  onClick?: (e: React.MouseEvent<HTMLButtonElement, MouseEvent>) => void;
+};
+export const UserAccountButton: React.FC<UserAccountButtonPropsType> = ({
```
skyboyer, line 12: ideally I'd suggest to extend `@material-ui` types

```
type UserAccountButtonPropsType = {
  type: 'activate' | 'deactivate' | 'create';
};

export const UserAccountButton: React.FC<UserAccountButtonPropsType & MaterialUIButtonProps> = ({
```

but we don't do this in this project. So let's keep current version.

src/old/lib/types.ts
```diff
@@ -1,9 +1,14 @@
-export enum ExpertStatus {
+export enum UserStatus {
   NEW = 'NEW',
   ACTIVE = 'ACTIVE',
   DELETED = 'DELETED',
 }
 
+export type UserStatusType =
```
skyboyer, line 7: btw in TypeScript 4.1 it could be done as

```
export type UserStatusType = `${UserStatus}`;
```

https://stackoverflow.com/questions/52393730/typescript-string-literal-union-type-from-enum/64966647#64966647

but we use 4.0.8 so I guess this is not supported

src/views/CreateUserAccountView/__tests__/CreateUserAccountPage.test.tsx
```diff
@@ -0,0 +1,126 @@
+import { act, render, screen } from '@testing-library/react';
+import React from 'react';
+import userEvent from '@testing-library/user-event';
+import CreateUserAccountPage from '../CreateUserAccountPage';
+
+jest.mock('react-router-dom', () => ({
```
skyboyer: I discourage you to mock router this way. Much more reliable way is to wrap component under test into `<MemoryRouter>`. Then we can mock initial history and check for navigation in stable way.

https://v5.reactrouter.com/web/guides/testing

as an additional side effect, you would not need to mock `useQuery` thing

src/views/CreateUserAccountView/__tests__/CreateUserAccountPage.test.tsx
```diff
@@ -0,0 +1,126 @@
...
+    };
+    return new URLSearchParams(mockLocation.search);
+  }),
+}));
+
+describe('Tests for CreateUserAccountPage', () => {
+  it('Should render CreateUserAccountPage', () => {
+    const { container } = render(<CreateUserAccountPage />);
+    expect(container.firstChild?.firstChild).toHaveClass(
+      'CreateUserAccountPage-FormBox-1',
```
skyboyer: instead of testing class names which are free to change any minute, we could test presence of the text inputs and buttons. Or with all the next test cases this test case "it renders" is redundant - if all next test cases pass then form for sure renders

src/views/CreateUserAccountView/__tests__/CreateUserAccountPage.test.tsx
```diff
@@ -0,0 +1,126 @@
...
+    );
+  });
+  it('Should be able to type into inputs', async () => {
+    render(<CreateUserAccountPage />);
+
+    const inputs = screen.queryAllByTestId('basic-input') as HTMLInputElement[];
```
skyboyer: may we add `data-testid` to the input in order to avoid using `[0]` in the tests? or to use `document.querySelector("[name=...` as the last resort

src/views/CreateUserAccountView/__tests__/CreateUserAccountPage.test.tsx
```diff
@@ -0,0 +1,126 @@
...
+    // eslint-disable-next-line @typescript-eslint/require-await
+    await act(async () => {
```
skyboyer: why do you declare function as `async` if you don't use `await`?

`// eslint-disable-next-line` should not be a regular practice, it loses its importance then

src/views/CreateUserAccountView/__tests__/CreateUserAccountPage.test.tsx
```diff
@@ -0,0 +1,126 @@
...
+    });
+    expect(inputs[1].value).toBe('12345');
+  });
+
+  it('Button should be disabled till correct passwords will be written in inputs', async () => {
+    render(<CreateUserAccountPage />);
+
+    const inputs = screen.queryAllByTestId('basic-input') as HTMLInputElement[];
+
+    const button = screen.getByTestId('basic-button');
```
skyboyer: can it be `getByRole('button', { name: '....` instead?

src/components/Form/Buttons/__tests__/UserAccountButton.test.tsx
```diff
@@ -0,0 +1,62 @@
...
+  it('Should handle onClick', () => {
+    const handleClick = jest.fn();
+    const { container } = render(
+      <UserAccountButton
+        type="activate"
+        label="Hello world"
+        onClick={handleClick}
+      />,
+    );
+    userEvent.click(container.firstChild as TargetElement);
```
skyboyer: `.firstChild` is veeeeeery unreliable selector. Probably, anything else is better than it.
`getByRole('button'`(if it works with material-ui's buttons), `getByText`, `.querySelector('button'` - here are much better candidates

src/components/Form/СreateUserAccountForm/styles/CreateUserAccountForm.style.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+export const useStyles = makeStyles(
+  (theme: Theme) => ({
+    FormBox: {
+      '& ': {
+        display: 'flex',
+        gap: '15px',
+      },
+      '& .MuiFormLabel-root': {
+        fontFamily: 'Raleway',
```
skyboyer: I see we are hard-coding font-family all the way across the project, but typically we should better specify it once to the root(e.g. for `<body>`) or at least to put it into `theme`

src/components/Form/СreateUserAccountForm/styles/CreateUserAccountForm.style.tsx
```diff
@@ -0,0 +1,51 @@
...
+        color: 'black',
+        fontWeight: 600,
+        fontSize: theme.spacing(3.2),
+      },
+      '& .MuiFormLabel-asterisk.MuiInputLabel-asterisk': {
```
skyboyer: I think this override should be elsewhere, not at the form-level. We want all our forms to have `*` of the same color. Maybe, creating `useStyles()` for `App` component would be a better place

src/components/Form/СreateUserAccountForm/styles/CreateUserAccountForm.style.tsx
```diff
@@ -0,0 +1,51 @@
...
+    FormBox: {
+      '& ': {
+        display: 'flex',
+        gap: '15px',
+      },
+      '& .MuiFormLabel-root': {
+        fontFamily: 'Raleway',
+        color: 'black',
+        fontWeight: 600,
+        fontSize: theme.spacing(3.2),
```
skyboyer: I see we set font-size this way just in few places. could you please give me more details what's its purpose/advantage over `px` notation?

src/components/Form/СreateUserAccountForm/styles/CreateUserAccountForm.style.tsx
```diff
@@ -0,0 +1,51 @@
+import { makeStyles, Theme } from '@material-ui/core';
+
+export const useStyles = makeStyles(
+  (theme: Theme) => ({
+    FormBox: {
+      '& ': {
```
skyboyer: but why do we need `'& '`?

src/old/lib/utilities/API/api.ts
```diff
@@ -370,3 +373,28 @@ export const updateAuthorById = async (
...
   return instance.put(`/author`, requestBody);
 };
+
+export const getAllEmails = async (): Promise<AxiosResponse<UserEmailType>> => {
+  return instance.get('/user/all-emails');
+};
+
+export const changeEnabledOfUser = async (requestBody: ChangeEnabledType) => {
+  // NEED CHANGE URL WHEN BACKEND UPDATES
+  return instance.put('/user/-1', requestBody);
```
skyboyer, line 383: will it work by now? or will it throw?

src/views/Profile/AdminTable/ActionButtons.tsx
```diff
@@ -137,7 +137,7 @@ const ActionButtons: React.FC<IActionButtons> = ({
       label: t(langTokens.admin.edit),
       modal: null,
       handler: () => {
-        history.push(editPostLink);
+        window.open(editPostLink);
```
skyboyer, line 140: I don't see it in https://github.com/ita-social-projects/dokazovi-requirements/issues/634
why it should be opened in new window?

src/components/Form/СreateUserAccountForm/CreateUserAccountForm.tsx
```diff
@@ -0,0 +1,108 @@
...
+    }
+    return errors;
+  }, [userEmail, visitFields, usersEmails]);
+
+  const handleSubmitForm = (
+    e: React.MouseEvent<HTMLButtonElement, MouseEvent>,
+  ) => {
+    onClick(e, userEmail);
+  };
+  console.log(isLoading);
```
skyboyer: oops, `console.log`

src/components/Form/СreateUserAccountForm/CreateUserAccountForm.tsx
```diff
@@ -0,0 +1,108 @@
...
+      errors.email = i18n.t(langTokens.admin.fieldCantBeEmpty);
+    } else if (userEmail && !regex.validEmail.test(userEmail)) {
+      errors.email = i18n.t(langTokens.admin.wrongEmail);
+    } else if (
+      userEmail &&
+      usersEmails.publicEmail.some((mail) => mail === userEmail)
+    ) {
+      errors.email = i18n.t(langTokens.admin.diffEmail);
+    } else if (!userEmail && !visitFields.email) {
+      errors.email = ' ';
```
skyboyer, line 61: so there is un-empty error message but it's not visible to the user?

src/views/CreateUserAccountView/CreateUserAccountPage.tsx
```diff
@@ -0,0 +1,144 @@
...
+
+const CreateUserAccountPage = () => {
+  const query = useQuery();
+  const history = useHistory();
+  const { t } = useTranslation();
+
+  const classes = useStyles();
+
+  // eslint-disable-next-line @typescript-eslint/unbound-method
+  const { register, handleSubmit, errors, watch, formState } = useForm<
```
skyboyer: what is ESLint error for `unbound-method` rule here?

src/views/CreateUserAccountView/CreateUserAccountPage.tsx
```diff
@@ -0,0 +1,144 @@
...
+        />
+      </Box>
+    </Container>
+  );
+
+  const createUserAccForm = (
+    <Container>
+      <form onSubmit={handleSubmit(onSubmit)} className={classes.FormBox}>
+        <Typography variant="h3" component="p">
+          {`${t(langTokens.passwordForms.createUserAccountTitle)}:`}
```
skyboyer: minor: punctuation should _always_ be a part of i18n strings; depending on the language it may appear to the left or to the right from "main" text

src/views/CreateUserAccountView/CreateUserAccountPage.tsx
```diff
@@ -0,0 +1,144 @@
...
+    }
+  };
+
+  const loadingView = (
+    <Box className="mainLoading">
+      <CircularProgress />
+    </Box>
+  );
+
+  const createUserAccSuccess = (
```
skyboyer: I suggest to make `createUserAccSuccess` and `createUserAccForm` to be separate components. So `CreateUserAccountForm` would have form and its state while `CreateUserAccountPage` would display either a form, or success message + button, or loading indicator

src/views/CreateUserAccountView/CreateUserAccountPage.tsx
```diff
@@ -0,0 +1,144 @@
...
+  const onSubmit = async (inputs: ICreateUserAccount) => {
+    const token = query.get('token');
+    if (!token) {
+      toast.error('No token provided');
+      return;
+    }
+    try {
+      setLoadingStatus(LoadingStatusEnum.pending);
+      await activateUser(token, inputs.newPassword, inputs.matchPassword);
+      toast.success('User was activated');
```
skyboyer: I think all the messages should come from i18n. Or is there a special reason why it cannot be done this way here?

src/components/Form/СreateUserAccountForm/CreateUserAccountForm.tsx
```diff
@@ -0,0 +1,108 @@
...
+type CreateUserAccountFormPropsType = {
+  usersEmails: UserEmailType;
+  onClick: (
+    e: React.MouseEvent<HTMLButtonElement, MouseEvent>,
+    email: string,
+  ) => void;
+  currentUserEmail?: string;
+  isLoading?: boolean;
+};
+type ErrorFieldsType = {
```
skyboyer, line 20: just a thought: I think we could use composite `StateType = { email: string }` so we could re-use it in `ErrorFieldsType` and `VisitedFieldsType` as mapped-type thing:

```
type ErrorFieldsType = {
    [property in keyof StateType]: string
};

type VisitedFieldsType = {
    [property in keyof StateType]: boolean
};
```

just to avoid hard-coding the same structure again and again. 
Moreover, we can specify type based on initial value:

```
const initialValues = {
  email: ''
};

type ErrorFieldsType = {
  [prop in keyof typeof initialState]: string
};

...
  ... =  useState({ ...initialState });
```

something like this. However, I'm not sure if it worth it here and I don't think this pattern is used anywhere else in the project

src/components/Form/СreateUserAccountForm/CreateUserAccountForm.tsx
```diff
@@ -0,0 +1,108 @@
...
+  ) => {
+    setUserEmail(e.target.value);
+  };
+
+  const errorMessage: ErrorFieldsType = useMemo(() => {
+    const errors: ErrorFieldsType = {
+      email: '',
+    };
+
+    if (currentUserEmail) {
```
skyboyer, line 49: could you please clarify this particular condition? so if prop `currentUserEmail` is passed from the outside, we immediately show the error message?

src/components/Form/СreateUserAccountForm/CreateUserAccountForm.tsx
```diff
@@ -0,0 +1,108 @@
...
+type VisitedFieldsType = {
+  email: boolean;
+};
+export const CreateUserAccountForm: React.FC<CreateUserAccountFormPropsType> = ({
+  onClick,
+  usersEmails,
+  currentUserEmail,
+  isLoading,
+}) => {
+  const [userEmail, setUserEmail] = useState('');
```
skyboyer, line 32: what if we use `react-hook-form` here? would code be less verbose?

src/views/Profile/PersonalInfo/PersonalInfo.tsx
```diff
@@ -249,189 +315,272 @@ export const PersonalInfo: React.FC<IEditAuthorProps> = ({
...
-          type="sign"
-          label={i18n.t(langTokens.common.acceptChanges)}
-          className={classes.BasicButton}
-          onClick={buttonClickHandler}
-        />
-      </Box>
-    </form>
+        </Box>
+      </form>
+      {isAdmin && author && !isCurrentUser && (
```
skyboyer: is it easy enough to move this part into separate component?

src/views/Profile/PersonalInfo/PersonalInfo.tsx
```diff
@@ -249,189 +315,272 @@ export const PersonalInfo: React.FC<IEditAuthorProps> = ({
...
+                    usersEmails={usersEmails}
+                    onClick={handleActivateUser}
+                    isLoading={isLoading}
+                  />
+                </Box>
+              )}
+            </>
+          )}
+          <Box className={classes.ButtonBox}>
+            {enabled && status === UserStatus.ACTIVE && (
```
skyboyer: having multiple complex mutually exclusive conditions with ` ... && ...` is really tricky and risky - we may end up with multiple conditions evaluated to `true` at some moment. another concern is we don't have default value(when neither of our conditions matches) - which may be desired or not, so is confusing to the reader.

so instead of putting this into JSX, we could calculate `buttonState` variable so it will be guaranteed only single block to be rendered. It may be done with 
```
let buttonState;
if (..) { buttonState = '...' }  // only single case will be applied
else if ( ... ) { buttonState = ' ...' ` 
else if ... 
else {
  buttonState = 'hidden'; // example of default case
}
```
or it can be done with a separate function with early exit pattern in it; this will allow us to stop checking for specific variables at some point because we will be sure it has some specific value:

```
function getButtonState({enabled, status, toggleButton}) {
  if (enabled && status === UserStatus.ACTIVE) {
    return 'deactivate';
  }
  if (enabled && status !== UserStatus.ACTIVE) {
    return 'hidden';
  }
 // from here and down `enabled === false` and we don't need to check it any more
  .....  
  return 'hidden'; // default case
}
```

this "skipping condition check" is not a main goal, but depending on specific scenario can heavily simplify the conditions tree

skyboyer (commented):
I like the changes! 
still, have some questions(where I did not get the requirements) and comments

src/components/Form/Buttons/styles/UserAccountButton.style.tsx
YuriyHryshchenko: I will change it for theme.pallete.warning

src/components/Form/Buttons/styles/UserAccountButton.style.tsx
YuriyHryshchenko: I just take color, which we use for create new author button, just to commit style, so i don`t think we need to change it

src/components/Form/Buttons/styles/UserAccountButton.style.tsx
YuriyHryshchenko: Okay

src/components/Form/Buttons/UserAccountButton.tsx
YuriyHryshchenko: Okay, thanks for advice

src/components/Form/СreateUserAccountForm/CreateUserAccountForm.tsx
YuriyHryshchenko: If user has currentEmail - it means that he had already sent his email to activate his acc. So in meeting, we discussed that should be displayed error in this case just to not allow user to send new email, because he need to confirm old

src/components/Form/СreateUserAccountForm/CreateUserAccountForm.tsx
YuriyHryshchenko: I think we can change it

src/old/lib/types.ts
YuriyHryshchenko: Thanks for advice

src/views/CreateUserAccountView/__tests__/CreateUserAccountPage.test.tsx
YuriyHryshchenko: Thanks, i didn`t know about MemoryRouter, its much more convenient.

src/views/CreateUserAccountView/__tests__/CreateUserAccountPage.test.tsx
YuriyHryshchenko: Ok, sure

src/views/CreateUserAccountView/__tests__/CreateUserAccountPage.test.tsx
YuriyHryshchenko: We can`t add testid because we have it already in our component, but i will change to document.querySelector

src/views/CreateUserAccountView/__tests__/CreateUserAccountPage.test.tsx
YuriyHryshchenko: Yes, I will change it

src/components/Form/Buttons/__tests__/UserAccountButton.test.tsx
YuriyHryshchenko: Okay, i will change it

src/views/Profile/AdminTable/ActionButtons.tsx
YuriyHryshchenko: We discussed it in meeting

src/components/Form/СreateUserAccountForm/CreateUserAccountForm.tsx
YuriyHryshchenko: I will remove

src/components/Form/СreateUserAccountForm/CreateUserAccountForm.tsx
YuriyHryshchenko: Just to disable button to show user that he need to input email

src/components/Form/СreateUserAccountForm/styles/CreateUserAccountForm.style.tsx
YuriyHryshchenko: I will change it globaly for Form label

src/components/Form/Buttons/styles/UserAccountButton.style.tsx
YuriyHryshchenko: We have different props on them. First prop is type -  for BasicButton its 'accept' or 'sign', for UserAccButton - 'activate', 'deactivate', 'create' - depending on this props we create styles. Also we have action prop on UserAccButton, but don`t have on BasicButton, depending on this prop we make type of button 'submit' or 'button'. I think I can add all this stuff to Basic button, and remove UserAccountButton component, but want to listen your opinion

src/components/Form/СreateUserAccountForm/styles/CreateUserAccountForm.style.tsx
YuriyHryshchenko: Ok, I will make it globally in theme styles

src/components/Form/СreateUserAccountForm/styles/CreateUserAccountForm.style.tsx
YuriyHryshchenko: I used, because saw we used it in some components. Its used for more consistency, scalability. If we change spacing in our theme we can change all fontSize across our components instead of going manually and changing px

src/components/Form/СreateUserAccountForm/styles/CreateUserAccountForm.style.tsx
YuriyHryshchenko: To target parent element

src/old/lib/utilities/API/api.ts
YuriyHryshchenko: As i know it throws still

src/views/CreateUserAccountView/CreateUserAccountPage.tsx
YuriyHryshchenko: ESLint: 
Avoid referencing unbound methods which may cause unintentional scoping of `this`.
If your function does not access `this`, you can annotate it with `this: void`, or consider using an arrow function instead.

Just make with example found somewhere in our code

src/views/CreateUserAccountView/CreateUserAccountPage.tsx
YuriyHryshchenko: I will change

src/views/CreateUserAccountView/CreateUserAccountPage.tsx
YuriyHryshchenko: Okay, I will remake it

src/views/CreateUserAccountView/CreateUserAccountPage.tsx
YuriyHryshchenko: No special reasons, I will rewrite this

src/components/Form/СreateUserAccountForm/CreateUserAccountForm.tsx
YuriyHryshchenko: I think for now it`s more simple to leave it as it is

src/views/Profile/PersonalInfo/PersonalInfo.tsx
YuriyHryshchenko: I will try

src/views/Profile/PersonalInfo/PersonalInfo.tsx
YuriyHryshchenko: Sorry, i don`t understand how i can apply this, i can do separate if statments with return jsx or explain me more about this technique in our case, please

skyboyer: I'm sorry for over detailed explanation, just want to give all the reasoning

So final goal is to make code which:

1. guarantees that no more than single piece of JSX will be rendered under any conditions
2. clearly outputs a default case when no condition matches - or explicitly shows that we don't need anything to be shown

Current version of 

```
{condition1 && condition2 && (JSX1)}
{condition3 && !condition2 && (JSX2)}
{/* etc */}
```

does not help with both. And I see next options here:

Option 1. Introducing single state-related variable right in the render logic:

```
let buttonTypeToShow: ActionButtonType = 'nothing';
if (condition1 && condition2) {
  buttonTypeToShow = 'deactivate';
} else if (condition1 && !condition2) {
  buttonTypeToShow = 'activate';
} else if (...) 
.....
  {buttonTypeToShow === 'activate' && (...)}
  {buttonTypeToShow === 'deactivate' && (....)}
```
It's clearly only single case satisfies and we can see that default value for `buttonTypeToShow` does not render anything - intentionally, not accidetnally.

Option 2. Introducing separate component with early exit pattern

```
function PersonalInfoActionButton({ condition1, condition2, condition3, onClick, children }) {
  if (condition1 && condition2) {
    return (JSX1);
  }
  if (condition1 && !condition2) {
    return (JSX2);
  }
  ....
 return null; // if nothing match we don't render anything
}
```

Option 3. The same as option 2 but without introducing separate component. IMO this is the worst way among those three - we cannot just read rendering part of `<PersonalInfo>` and have to navigate back and forth across the functions to understand what will be final result. But still, it saves us from case when mutliple conditions matches and few different buttons are rendered side-by-side:

```
function PersonalInfo({...}) {
  const renderButton = () => {
    if (condition1 && condition2) { // 
      return (JSX1);
    }
    if (condition1 && !condition2) {
      return (JSX2);
    }
    ....
   return null; // if nothing match we don't render anything
  }

 ....
  return (
  ....
    {renderButton()}
  ....
}
```

src/components/Form/СreateUserAccountForm/CreateUserAccountForm.tsx
skyboyer: with current version of code `initialValues` constant is nowhere used. And I guess it will be more confusing to the future reader than having hard-coded types here and there(I mean `email: string`, `email: boolean` in initial version)

let's revert to initial version. Approach I proposed might make sense with many fields, but does not align with single `useState(primitiveStringValue)`

src/components/Form/СreateUserAccountForm/styles/CreateUserAccountForm.style.tsx
skyboyer: I think that `&` targets element itself(here it's any element with classname of `props.classes.FormBox`)

My assumption that

```
FormBox: {
  display: 'flex',
  gap: '15px',
  '& ,MuiFormLabel-root': {
    color: 'black',
    fontWeight: 600,
```

will work the same but without extra nesting in `'& '`. Or am I wrong?

src/components/Form/СreateUserAccountForm/styles/CreateUserAccountForm.style.tsx
skyboyer: I see what you mean. Reusability is typically a good thing.

However, here I find this fragile. Normally, paddings and font-size are mutually independent - you may have small text with large gaps, you may have large text with smaller gaps. Making font-size depending on default padding value leads for us to be unsure whether changing it will not make some parts unreadable at all.

Also, it's inconsistent: we already have `fontSize` as `theme.spacing(3.6)`, `theme.spacing(2.8)`, `theme.spacing(6.8)`, and also we have hard-coded values of 11, 14, 17, 18, 19, 20 and even more options - all in pixels.
 
In long run, I would suggest to introduce extra theme values with semantic names. E.g. `theme.fontSize.normal` and `theme.fontSize.heading1` or alike.

But in short run, and taking into account existing code, I suggest to use hard-coded value in `px`

src/components/Posts/PostInfo/PostInfo.styles.ts
```diff
@@ -27,7 +27,7 @@ export const useStyles = makeStyles(
       '& > li': {
         padding: '5px 12px',
         color: theme.palette.common.white,
-        backgroundColor: theme.palette.custom.colorAquaBlue,
+        backgroundColor: theme.palette.custom.main,
```
skyboyer, line 30: nice!

src/styles/theme/index.ts
```diff
@@ -29,6 +29,14 @@ export const MAIN_THEME = createTheme({
         fontFamily: 'Raleway',
       },
     },
+    MuiFormLabel: {
```
skyboyer, line 32: nice!

src/styles/theme/palette.ts
```diff
@@ -36,6 +35,7 @@ export const PALETTE: PaletteOptions = {
   },
   divider: '#000',
   custom: {
-    colorAquaBlue: '#73ddff',
+    main: '#73ddff',
```
skyboyer, line 38: in the future if we have more colors here - it may make sense to give it some meaningful name. But so far it's good with `custom` too.

src/views/CreateUserAccountView/__tests__/CreateUserAccountPage.test.tsx
```diff
@@ -0,0 +1,146 @@
+import { act, render, screen } from '@testing-library/react';
+import React from 'react';
+import userEvent from '@testing-library/user-event';
+import { MemoryRouter } from 'react-router-dom';
+import CreateUserAccountPage from '../CreateUserAccountPage';
+
+// jest.mock('react-router-dom', () => ({
```
skyboyer: could you remove those commented lines? :)

src/components/Form/СreateUserAccountForm/styles/CreateUserAccountForm.style.tsx
YuriyHryshchenko: Yes, you are right

src/components/Form/СreateUserAccountForm/styles/CreateUserAccountForm.style.tsx
YuriyHryshchenko: Okay, will do fix px value

skyboyer (approved):
I think there might be some better(from maintainability standpoint) way to make forms - with smaller functions and less complex structure. But I'm unable to come with complete plan on this 🤷 

so let's go with those changes, overall look good to me

